PR #1, opened 2015-05-21
First draft

krl:
I would like it to actually do init automatically as well, then i'll update the readme.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,12 +3,15 @@ var _ = require('lodash')
 var Q = require('kew')
 var ipfs = require('ipfs-api')
 var waterfall = require('promise-waterfall')
+var freeport = require('freeport')
+var shutdown = require('shutdown-handler')
+var rimraf = require('rimraf')
 
 function configureNode (node, conf, cb) {
   waterfall(_.map(conf, function (value, key) {
     return function () {
       var def = Q.defer()
-      run('ipfs', ['config', key, value])
+      run('ipfs', ['config', key, value], {env: node.env})
         .on('error', function (err) { cb(err) })
         .on('end', function () { def.resolve() })
       return def.promise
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -0,0 +1,80 @@
...
+function configureNode (node, conf, cb) {
+  var delay = 0
+  var promises = _.map(conf, function (value, key) {
+    var def = Q.defer()
+    setTimeout(function () {
+      run('ipfs', ['config', key, value])
+        .on('error', function (err) { cb(err) })
+        .on('end', function () { def.resolve(true) })
+    }, delay * 200)
+    delay++
```
jbenet: may want to say why it has to run this way?

package.json
```diff
@@ -4,19 +4,32 @@
   "description": "simple controls for an ipfs node",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "mocha test"
```
travisperson: Missing comma.

package.json
```diff
@@ -4,19 +4,32 @@
...
+    "lint": "git diff --name-only --cached --relative | egrep .js$ | xargs --no-run-if-empty standard"
   },
+  "pre-commit": [
+    "test"
+  ]
```
travisperson: Missing comma.

krl:
test

jbenet:
nice screenshot

jbenet:
@krl this is great

index.js
krl: It seems there can be a race condition when you're setting two config variables at the same time. I will make an issue of this in my next sitting.

travisperson: Sounds like a good use case for waterfall promises?

test/test.js
```diff
@@ -0,0 +1,160 @@
+var run = require('comandante')
+var ipfsd = require('../index.js')
+var assert = require('assert')
+var fs = require('fs')
+var rimraf = require('rimraf')
+
+// happy lint
+// var it, before, after
```
travisperson: Does this actually work to allow standard to not complain about undefined? I usually see something like:
```
/* global beforeEach, describe, it */
```

index.js
travisperson: See: https://www.npmjs.com/package/promise-waterfall

krl:
@travisperson thank you for pointing me at waterfall promises, will def be using this from now on!

Pushed new version with this method

bcomnes:
:+1: Thanks for doing this!

bcomnes:
What is the relationship between https://www.npmjs.com/package/ipfs-api and https://www.npmjs.com/package/go-ipfs?

jbenet:
@bcomnes ipfs-api is a js lib for consuming/using the ipfs api provided by go-ipfs. (it's an http server run by go-ipfs, and other implementations should probably support it (match the api) too)

jbenet:
This LGTM

jbenet:
@krl shall we merge this?

bcomnes:
Baller! Nice work @krl   Just had a chance to play around with it.  Seems to work as expected

PR #2, opened 2015-06-05
patched shutdown-handler, fixed versions

krl:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ipfsd-ctl",
-  "version": "0.2.1",
+  "version": "0.2.3",
```
jbenet: shouldn't update this until after merge-- otherwise published versions would contain different things. published versions should be published from master

PR #5, opened 2015-06-10
Shot in the dark

harlantwood:
This fixed failures on my machine.

### DIFF
--- a/index.js
+++ b/index.js
@@ -185,7 +185,7 @@ module.exports = {
     }
     freeport(function (err, port) {
       if (err) return cb(err)
-      opts['Addresses.Gateway'] = '""'
+      opts['Addresses.Gateway'] = ''
       opts['Addresses.API'] = '/ip4/127.0.0.1/tcp/' + port
       var node = new Node(tempDir(), opts, true)
       node.init(function (err, newnode) {


index.js
```diff
@@ -185,7 +185,7 @@ module.exports = {
     }
     freeport(function (err, port) {
       if (err) return cb(err)
-      opts['Addresses.Gateway'] = '""'
```
jbenet: that's very odd. wonder if it was here for a reason. @krl @travisperson ?

travisperson: Not that I'm aware of.

krl:
Hmm, this was to not have the disposable nodes setup gateways

harlantwood:
I don't claim to understand the nuances here.  But I do know the tests were failing before this commit, and passing after it: https://circleci.com/gh/harlantwood/node-ipfsd-ctl

whyrusleeping:
@krl why was it being set to double empty quotes?

krl:
Jeromy Johnson <notifications@github.com> writes:

> @krl why was it being set to double empty quotes?
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/ipfs/node-ipfsd-ctl/pull/5#issuecomment-112560667

It was about not being able to pass an empty string to the config, in
the way it was written before.

PR #6, opened 2015-06-12
Make ipfsd.local return a node instance

krl:
This should be a minor version bump

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,26 +69,25 @@ var Node = function (path, opts, disposable) {
     },
     daemon: function (cb) {
       var t = this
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -69,26 +69,25 @@ var Node = function (path, opts, disposable) {
     },
     daemon: function (cb) {
       var t = this
-      var running = run(IPFS_EXEC, ['daemon', '--unrestricted-api'], {env: t.env})
```
jbenet: whoops, i didnt catch this. best prob not to publish this type of thing.

PR #7, opened 2015-06-16
add test coverage (process killing, local node setup and init)

krl:


jbenet:
I'm not sure using `tcp/0` is the right thing. I think we should be trying to prefer the standard ports? or let people _easily_ specify them? 

I'm thinking of cases where people for whatever reason need to open up a port. if `ipfsd-ctl` is getting used in things like electron-app and so on, people without any CLI experience may end up as users (there's a big difference between developers and non-developers in ability to change the port). Their UX may suffer if we're not careful about port numbers. particularly the API-- a bunch of things assume the API is at a well known port. (they shouldn't but im not sure what all needs to change at this point).

krl:
jbenet: ok, tcp/0 is only for the disposable test nodes, what electron-app uses is `ipfsd..local`

Here additionally, we have a grace period of 2 secons, which might be too short? before an additional SIGKILL is sent to the process.

index.js
```diff
@@ -3,12 +3,12 @@ var _ = require('lodash')
...
 var ipfs = require('ipfs-api')
 var waterfall = require('promise-waterfall')
-var freeport = require('freeport')
-var shutdown = require('shutdown-handler-krlpatch')
+var shutdown = require('shutdown-handler')
 var rimraf = require('rimraf')
 var fs = require('fs')
 
 var IPFS_EXEC = __dirname + '/node_modules/.bin/ipfs'
+var GRACE_PERIOD = 2000 // amount of ms to wait before sigkill
```
jbenet: yeah i think something like 5s or 10s more reasonable. if we can hide things (i.e. the electron-app doesnt hang while closing, but "appears to close" quickly, its fine if it finishes tearing down in the background.

index.js
```diff
@@ -67,28 +67,46 @@ var Node = function (path, opts, disposable) {
...
-        running
+        if (err) return cb(err)
+
+        t.subprocess = run(IPFS_EXEC, ['daemon'], {env: t.env})
           .on('error', function (err) {
             if ((err + '').match('daemon is running')) {
               // we're good
               cb(null, ipfs(conf.Addresses.API))
+            } else if ((err + '').match('non-zero exit code')) {
+              // ignore when kill -9'd
```
jbenet: how come this error happens on `startDaemon`? (Am missing something)

index.js
```diff
@@ -67,28 +67,46 @@ var Node = function (path, opts, disposable) {
...
               cb(err)
             }
           })
           .on('data', function (data) {
-            var match = (data + '').trim().match(/API server listening on/)
+            var match = (data + '').trim().match(/API server listening on (.*)/)
             if (match) {
-              cb(null, ipfs(conf.Addresses.API))
+              t.apiAddr = match[1]
+              cb(null, ipfs(t.apiAddr))
```
jbenet: :+1:

index.js
```diff
@@ -67,28 +67,46 @@ var Node = function (path, opts, disposable) {
...
             }
           })
       })
     },
+    stopDaemon: function () {
+      var t = this
+      var sp = t.subprocess
+
+      if (sp) {
+        sp.kill('SIGHUP')
```
jbenet: maybe `SIGTERM` ?

index.js
```diff
@@ -67,28 +67,46 @@ var Node = function (path, opts, disposable) {
...
+        setTimeout(function () {
+          sp.kill('SIGKILL')
```
jbenet: does it need to check? i guess not because this will just fail after.

ideally -- if things exit sooner -- we'd exit sooner. maybe the thing to do is a loop that checks whether the process has died every ~300ms and exits early if so?

index.js
krl: it happens when you kill -9, and the process exits non 0

index.js
krl: The problem with this is that i don't think there's a platform-independant way of doing this. For now i moved this into the tests, which we can run on platforms supporting 'kill' pid. No idea on how to do this in windows though.

I might be missing something though. But the sp.kill() here for example, returns immediatly, and you kind of don't know if it worked, just if it could find the pid or not.

index.js
krl: you are probably right

index.js
jbenet: something like:

```js
var closed = false
sp.on('close', function(code) {
  closed = true
})


sp.kill('SIGTERM')

var start = Date.now()
var killProc = function killProc() {
  if (closed == true) return // done

  var elapsed = Date.now() - start
  if (elapsed > GRACE_PERIOD) {
    sp.kill('SIGKILL') // done
  } else {
    setTimeout(killProc, 300)
  }
}
killProc()
```

test/test.js
```diff
@@ -89,3 +93,66 @@ describe('disposableApi node', function () {
...
+            done()
+          })
+      }, 100)
+    })
+
+    it('should be stopped', function () {
+      assert(!node.daemonPid())
+    })
+  })
+})
```
jbenet: may also want to test the API url grabbing and so on.

index.js
jbenet: ah ok

test/test.js
krl: i'll make that another issue/pr

krl:
@jebenet @travisperson This has grown a bit, but has better test coverage now. Should i squash this together?

index.js
```diff
@@ -67,28 +84,52 @@ var Node = function (path, opts, disposable) {
...
+
+      sp.on('close', function () {
+        clearTimeout(timeout)
+        cb && cb(null)
+      })
+
+      this.subprocess = null
+    },
+    daemonPid: function () {
+      return this.subprocess && this.subprocess.pid
```
travisperson: This will return a boolean. Did you want to actually return the pid?

travisperson:
I like this :+1:

index.js
krl: $ node
> 2 && 3
3

jbenet: this is where i like the coffeescript `thing?.anotherthing`

jbenet:
@krl 

> This has grown a bit, but has better test coverage now. Should i squash this together?

the commits seem to all pass (did you use [gpe](https://github.com/jbenet/git-push-each)? :) ): 



so it's up to you if we squash. (as long as they (a) all pass tests, (b) are reasonably "logically atomic") i dont care much.

jbenet:
I'll just merge it.

PR #14, opened 2015-09-18
Determines the IPFS binary path dynamically.

### LINKED ISSUE #13: Hardcoded IPFS_EXEC path not always correct
The current version of `ipfsd-ctl` hardcodes the following path:

`var IPFS_EXEC = __dirname + '/node_modules/.bin/ipfs'

This works fine when the module is at the root (e.g. git clone this repo and run `npm test`), but doesn't work when `ipfsd-ctl` is used as a dependency. In this `IPFS_EXEC` evaluates to 

`node_modules/ipfsd-ctl/node_modules/.bin/ipfs`

however, if the root module depends on `go-ipfs` the proper path to the binary will actually be `node_modules/.bin/ipfs`.

I'm preparing a PR that will use `require-resolve` to get the proper module path and look into its `/bin` directory.

noffle:
This makes the path valid regardless of whether 'ipfs-api' is a child module of this module, or of one of this module's parents.

Resolves #13.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,7 +9,7 @@ var rimraf = require('rimraf')
 var fs = require('fs')
 var path = require('path')
 
-var IPFS_EXEC = path.join(requireResolve('go-ipfs').pkg.root, '/bin/ipfs'))
+var IPFS_EXEC = path.join(requireResolve('go-ipfs').pkg.root, '/bin/ipfs')
 var GRACE_PERIOD = 7500 // amount of ms to wait before sigkill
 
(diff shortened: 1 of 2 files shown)


jbenet:
this LGTM. @diasdavid ?

bcomnes:
I like it.

jbenet:
published `ipfsd-ctl@0.3.4`

jbenet:
thanks @noffle!

Issue #15, opened 2015-09-18
Expose the API daemon's host/port


noffle:
It would be great if the resultant host and port were exposed by the resultant `disposableApi` object, so that other modules using `ipfs-api` could make use of the disposable daemon.

jbenet:
indeed :+1:

PR #17, opened 2015-09-18
Exposes a disposableApi's host and port.

noffle:
Addresses #15.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ var _ = require('lodash')
 var Q = require('kew')
 var ipfs = require('ipfs-api')
 var requireResolve = require('require-resolve')
+var multiaddr = require('multiaddr')
 var waterfall = require('promise-waterfall')
 var shutdown = require('shutdown-handler')
 var rimraf = require('rimraf')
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -109,7 +110,11 @@ var Node = function (path, opts, disposable) {
             var match = (data + '').trim().match(/API server listening on (.*)/)
             if (match) {
               t.apiAddr = match[1]
-              cb(null, ipfs(t.apiAddr))
+              var addr = multiaddr(t.apiAddr).nodeAddress()
+              var api = ipfs(t.apiAddr)
+              api.host = addr.address
+              api.port = addr.port
```
jbenet: this should probably be `api.apiAddr = t.apiAddr` because:
- want to stick to multiaddrs. this might not always be HTTP/TCP
- want to be explicit it's the addr of the api, because:

```
ipfs.host
ipfs.addr
```
might imply the node's actual (api protocol) address, instead of the api addr.

noffle:
That makes good sense: I've made the change!

jbenet:
thanks! published as `0.4.0`

PR #21, opened 2015-09-23
Require "go-ipfs" directly to get exe path.

bcomnes:
I believe this should normalize behavior between npm2 and 3 once the upstream path export is fixed: https://github.com/ipfs/npm-go-ipfs/pull/2

Closes https://github.com/ipfs/node-ipfsd-ctl/issues/20


bcomnes:
Ok I think this is good to go now.  What happens when we re-run the tests?

diasdavid:
Nice! tested this and also tested this new ipfsd-ctl on node-ipfs-api and every test passes, should solve the issue for everyone, including https://github.com/ipfs/node-ipfs-api/pull/64

PR #22, opened 2015-09-25
Update package.json to use semver ranges

bcomnes:
Maybe there is a reason you are not using semver ranges on the dependencies, but this patch fixes that!  Feel free to reject if this isn't wanted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,17 +22,17 @@
   "license": "MIT",
   "dependencies": {
     "go-ipfs": "0.3.7-3",
-    "ipfs-api": "2.2.1",
-    "kew": "0.5.0",
-    "lodash": "3.6.0",
+    "ipfs-api": "^2.2.1",
+    "kew": "0.7.0",
+    "lodash": "^3.6.0",
     "multiaddr": "^1.0.0",
     "promise-waterfall": "0.1.0",
-    "rimraf": "2.3.4",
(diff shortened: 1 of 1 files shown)


bcomnes:
Oops, got ahead of myself here.  This will pass once its rebased or merged with #21

package.json
```diff
@@ -21,19 +21,18 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "go-ipfs": "0.3.6",
-    "ipfs-api": "2.2.1",
-    "kew": "0.5.0",
-    "lodash": "3.6.0",
+    "go-ipfs": "0.3.7-3",
```
jbenet: do we need to be specific? can we not say `0.3.7` and get the (only) one that's published? i checked this manually with `npm install 0.3.7` and it installed one... but not with a dep)

jbenet:
ok maybe rebase and force push?

bcomnes:
I'll rebase this evening after work.

package.json
bcomnes: I'm not sure but I'll fix it if it works. I'll I've to check the semver calculator when I get home. 

--
Bret Comnes
http://bret.io

diasdavid:
LGTM :) thanks @bcomnes !

Issue #24, opened 2015-10-21
Enable starting a Node without using the bootstrap peers nor mdns (for controlled tests)


diasdavid:

diasdavid:
important to note: 
```
15:40 <+daviddias> whyrusleeping: does go-ipfs have a command to start a daemon and avoid using mdns and bootstrap nodes or is that all env vars?
15:42 <@whyrusleeping> daviddias: uhm....
15:42 <@whyrusleeping> we were going to have a setting so you could pass arbitrary config values to init
15:42 <@whyrusleeping> but i dont know if that ever happened
15:43 <@whyrusleeping> you'll have to run init, make the changes, and then run the daemon
```

diasdavid:
done on #25

jbenet:
maybe would be good to add  `--discovery none` and `--routing none` options

jbenet:
cc @whyrusleeping

PR #25, opened 2015-10-22
add setConfig feature and test

diasdavid:
#24 

@Dignifiedquire @VictorBjelkholm could you review this?

### DIFF
--- a/index.js
+++ b/index.js
@@ -138,13 +138,26 @@ var Node = function (path, opts, disposable) {
     daemonPid: function () {
       return this.subprocess && this.subprocess.pid
     },
-    getConf: function (key, cb) {
+    getConfig: function (key, cb) {
+      if (typeof key === 'function') {
+        cb = key
+        key = ''
+      }
       var t = this
       var result = ''
       run(IPFS_EXEC, ['config', key], {env: t.env})
         .on('error', cb)
         .on('data', function (data) { result += data })
-        .on('end', function () { cb(null, result.trim()) })
+        .on('end', function () {
+          cb(null, result.trim())
+        })
+    },
+    setConfig: function (key, value, cb) {
+      var t = this
+      run(IPFS_EXEC, ['config', key, value, '--json'], {env: t.env})
+        .on('error', cb)
+        .on('data', function (data) {})
+        .on('end', function () { cb() })
     },
     replaceConf: function (file, cb) {
       var t = this
--- a/test/test.js
+++ b/test/test.js
@@ -226,3 +226,45 @@ describe('setting up an initializing a local node', function () {
     })
   })
 })
+
+describe('change config values of a disposable node', function () {
+  this.timeout(20000)
+
+  var ipfsNode
+
+  before(function (done) {
+    ipfsd.disposable(function (err, node) {
+      if (err) {
+        throw err
+      }
+      ipfsNode = node
+      done()
+    })
+  })
+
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -138,13 +138,26 @@ var Node = function (path, opts, disposable) {
     daemonPid: function () {
       return this.subprocess && this.subprocess.pid
     },
-    getConf: function (key, cb) {
+    getConfig: function (key, cb) {
+      if (typeof key === 'function') {
```
whyrusleeping: what kind of weird language uses three equal signs?

diasdavid: For the same reason you some times ask a person 'are you sure' and other times 'are you really sure' :D

index.js
```diff
@@ -138,13 +138,26 @@ var Node = function (path, opts, disposable) {
...
       run(IPFS_EXEC, ['config', key], {env: t.env})
         .on('error', cb)
         .on('data', function (data) { result += data })
-        .on('end', function () { cb(null, result.trim()) })
+        .on('end', function () {
+          cb(null, result.trim())
+        })
+    },
+    setConfig: function (key, value, cb) {
+      var t = this
```
whyrusleeping: do we use es6 (or whatever) that lets you use `let` ?

test/test.js
```diff
@@ -226,3 +226,45 @@ describe('setting up an initializing a local node', function () {
     })
   })
 })
+
+describe('change config values of a disposable node', function () {
+  this.timeout(20000)
```
whyrusleeping: random arbitrary number? maybe define it somewhere and explain why the value was chosen?

index.js
VictorBjelkholm: Hm, `package.json` doesn't contain any node version (somebody should open an issue about that) so not sure what compatibility we're reaching for. But with that said, if we were using ES2015 (new name for ES6, yay :( ), this assignment wouldn't have been there since the functions below would have been [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions).

test/test.js
VictorBjelkholm: Basically, it's chosen because IPFS usually responds within those timeframes. In node-ipfs-api we have to do the same, since the response times varies... Kind of sucks

VictorBjelkholm:
Looking all good to me!

index.js
```diff
@@ -138,13 +138,26 @@ var Node = function (path, opts, disposable) {
...
+        key = ''
+      }
       var t = this
       var result = ''
       run(IPFS_EXEC, ['config', key], {env: t.env})
         .on('error', cb)
         .on('data', function (data) { result += data })
-        .on('end', function () { cb(null, result.trim()) })
+        .on('end', function () {
+          cb(null, result.trim())
```
Dignifiedquire: It would be nice if the result gets parsed as we are passing and retriving json right?

diasdavid: that would be nice, but also break current expectations. having both as a option would be ideal

index.js
diasdavid: I'm interested to learn what you would recommend for us to adopt from ES6 throughout our Node.js stuff. 

Why would var self = this be an anti-pattern? I know that now with fat arrows, we can avoid having to do that, but that can be more of a style preference than actually an anti-pattern, right?

diasdavid:
Thanks for CR'ing this :)

PR #65, opened 2016-04-09
Update ipfs-api package version...

haadcode:
...and add a test to make sure we're using the right version of ipfs-api.

The module is currently **BROKEN** as it's using a different version of ipfs-api: 
*Adding a directory recursively returns a different hash every time. This seems to be due to ipfs-api version mismatch (2.10.x vs. 3.0.0). This pull request will update the ipfs-api package version and fix the problem (verified by the test in the PR).*

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "ipfs-api": "^2.10.1",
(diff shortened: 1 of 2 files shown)


diasdavid:
That is very weird. Does that mean that ipfs-api `2.10` was returning different hashes?

PR #66, opened 2016-04-11
Fix default IPFS exec path when used via Electron

haadcode:
process.cwd() in Electron will always return '/' instead of the current directory (https://github.com/electron/electron/issues/2108). By checking if we're using Electron, we can get Electron app's default folder instead of '/'.

```javascript
const rootPath = process.versions['electron'] ? process.resourcesPath + '/app/' : process.cwd()
const ipfsDefaultPath = path.join(rootPath, 'node_modules/go-ipfs-dep/go-ipfs/ipfs')
```

This PR adds that check and tests to cover the Electron vs Node.js use cases.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -10,11 +10,12 @@ const path = require('path')
 
 // this comment is used by mocha, do not delete
 /*global describe, before, it*/
+
 describe('ipfs executable path', function () {
   this.timeout(2000)
   let Node
 
-  it('has the correct path when used via Electron', (done) => {
+  it('has the correct path when used via Electron', () => {
     process.versions['electron'] = '0.0.0-test'   // Electron sets its version to the array --> we know that we're using Electron
     process.resourcesPath = '/test/path/one/more' // Path to the Electron app, set by Electron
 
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -10,11 +10,41 @@ const path = require('path')
...
+    process.versions['electron'] = '0.0.0-test'   // Electron sets its version to the array --> we know that we're using Electron
+    process.resourcesPath = '/test/path/one/more' // Path to the Electron app, set by Electron
+
+    // Force reload of the module (pathing is handled globally in ../lib/node.js)
+    delete require.cache[require.resolve('../lib/node.js')]
+    Node = require('../lib/node.js')
+
+    var node = new Node()
+    assert.equal(node.exec, path.join(process.resourcesPath, '/app', 'node_modules/go-ipfs-dep/go-ipfs/ipfs'))
+    done()
```
dignifiedquire: Not an async test, so no need to use `done`

test/test.js
```diff
@@ -10,11 +10,41 @@ const path = require('path')
...
+    delete process.versions['electron']
+    delete process.resourcesPath
+
+    // Force reload of the module (pathing is handled globally in ../lib/node.js)
+    delete require.cache[require.resolve('../lib/node.js')]
+    Node = require('../lib/node.js')
+
+    var node = new Node()
+    assert.equal(node.exec, path.join(process.cwd(), 'node_modules/go-ipfs-dep/go-ipfs/ipfs'))
+    done()
```
dignifiedquire: Not an async test, so no need to use `done`

diasdavid:
LGTM :)

PR #67, opened 2016-04-22
add check for npm2 path and remove electron path check

FrauBienenstich:
this pr will fix the problem of ipfsd-ctl not working with npm2. in addition the electron path check is not needed anymore because of the use of __dirname in rootPath.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -6,6 +6,7 @@ const ipfsApi = require('ipfs-api')
 const run = require('subcomandante')
 const fs = require('fs')
 const rimraf = require('rimraf')
+const mkdirp = require('mkdirp')
 const path = require('path')
 
 // this comment is used by mocha, do not delete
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -15,28 +16,34 @@ describe('ipfs executable path', function () {
...
-    Node = require('../lib/node.js')
-
-    var node = new Node()
-    assert.equal(node.exec, path.join(process.cwd(), 'node_modules/go-ipfs-dep/go-ipfs/ipfs'))
+  it('has the correct path when installed with npm2', (done) => {
+    process.env.testpath = '/tmp/ipfsd-ctl-test/node_modules/ipfsd-ctl/lib' // fake __dirname
+    let npm2Path = '/tmp/ipfsd-ctl-test/node_modules/ipfsd-ctl/node_modules/go-ipfs-dep/go-ipfs'
+
+    mkdirp(npm2Path, (err) => {
+      if (err) console.log(err)
```
diasdavid: as a standard in all the code we do, we put curlies on all if clauses

diasdavid:
This is great @FrauBienenstich ! Had just a small nitpick codewise, but otherwise LGTM

I could not infer, however, if this would also fix the problem if ipfsd-ctl is a dependency of a dependency for the main project (and so forth for N levels of nestness). Have you tested it? 

Thank you :)

diasdavid:
Another possible fix for the issue, is make go-ipfs export a function with its __dirname, so that other modules can always tell where it is, not needing to make clauses for npm2 or npm3

FrauBienenstich:
nitpicks are appreciated! :)
this also works as a dependency of a dependency, tested it on three levels of dependencies (i had some help with that from @haadcode)

diasdavid:
awesome! Let's get this merged then :)

FrauBienenstich:
yay! thank you! :)

harlantwood:
Under both npm 2 and npm 3, I believe you can use:

`node_modules/.bin/ipfs`

haadcode:
> Under both npm 2 and npm 3, I believe you can use:
> `node_modules/.bin/ipfs`

the ipfs binary doesn't get copied to `node_modules/.bin` in either case.

diasdavid:
Actually, it is also an idea. If we point in the 'bin' key of package json to the binary, then we can always find it on node_modules/.bin/ipfs

haadcode:
> Actually, it is also an idea. If we point in the 'bin' key of package json to the binary, then we can always find it on node_modules/.bin/ipfs

If you change this, make sure to put the Electron check back in. The upside of using this logic (and using __dirname) is that it works in Electron too, whereas process.cwd() returns "/" (as in filesystem root) in Electron.

PR #89, opened 2016-07-13
New API + use bin-wrapper (bundle in all arch without any concern) + global install

### LINKED ISSUE #78: Upgrade js-ipfsd-ctl to use bin-wrapper
Our friends from Akasha made a great wrapper around the IPFS binary (https://github.com/AkashaProject/ipfs-connector/blob/master/src/IpfsBin.ts) that is way more portable when our current approach in js-ipfsd-ctl. 

It uses https://www.npmjs.com/package/bin-wrapper which enables the app to select the binary for the right arch seamlessly (a issue we were hitting on electron).

Let's move to the bin-wrapper approach :)

JGAntunes:
[WIP]
This should close #78, it's still a work in progress but let me know what you think so far.

There are still some issues though and further testing is needed. Also somehow the ipfs init command hangs for version 0.4.2, still have to look at what's going on with it.


JGAntunes:
Ok, looks like the 0.4.2 issue is a known one https://github.com/ipfs/go-ipfs/issues/2748

package.json
```diff
@@ -43,13 +43,13 @@
...
   "license": "MIT",
   "dependencies": {
-    "go-ipfs-dep": "0.4.1",
     "ipfs-api": "^4.1.0",
     "multiaddr": "^2.0.0",
     "rimraf": "^2.4.5",
     "run-series": "^1.1.4",
     "shutdown": "^0.2.4",
-    "subcomandante": "^1.0.5"
+    "subcomandante": "^1.0.5",
```
diasdavid: do we still need this module with bin-wrapper?

src/node.js
```diff
@@ -10,187 +10,218 @@ const shutdown = require('shutdown')
...
 
-const rootPath = process.env.testpath ? process.env.testpath : __dirname
-
-const ipfsDefaultPath = findIpfsExecutable()
+const BinWrapper = require('bin-wrapper')
+
+// TODO maybe this should go to some config file
+const DIST_BASE = 'https://dist.ipfs.io/go-ipfs'
+const VERSION = 'v0.4.1'
+const PROJECT_BASE = DIST_BASE + '/' + VERSION + '/go-ipfs_' + VERSION + '_'
```
diasdavid: Yeah, moving it to a config sounds good :)

src/node.js
```diff
@@ -10,187 +10,218 @@ const shutdown = require('shutdown')
...
+    .src(PROJECT_BASE + 'darwin-386.tar.gz', 'darwin', 'ia32')
+    .src(PROJECT_BASE + 'darwin-amd64.tar.gz', 'darwin', 'x64')
+    .src(PROJECT_BASE + 'freebsd-amd64.tar.gz', 'freebsd', 'x64')
+    .src(PROJECT_BASE + 'linux-386.tar.gz', 'linux', 'ia32')
+    .src(PROJECT_BASE + 'linux-amd64.tar.gz', 'linux', 'x64')
+    .src(PROJECT_BASE + 'linux-arm.tar.gz', 'linux', 'arm')
+    .src(PROJECT_BASE + 'windows-386.tar.gz', 'win32', 'ia32')
+    .src(PROJECT_BASE + 'windows-amd64.tar.gz', 'win32', 'x64')
+    .use(process.platform === 'win32' ? 'ipfs.exe' : 'ipfs')
+    // .version('>=1.71');
```
diasdavid: what's this line?

diasdavid:
This is awesome @JGAntunes :D 

Yeah, the issue with `0.4.2` exists, it has been fixed for 0.4.3, we just have to wait for that release. This module hasn't have any documentation so far, would you like to give a stab at it too? It will help us make sure we are fulfilling the right interface needs, @haadcode and @dignifiedquire will probably want to review the interface too.

test/index.spec.js
```diff
@@ -322,7 +322,7 @@ describe('external ipfs binaray', () => {
     ipfsd.local((err, node) => {
       if (err) throw err
 
-      assert.equal(node.exec, '/some/path')
+      assert.equal(node.exec, '/some/path/ipfs')
```
diasdavid: Could you change from `assert` to `expect`?

test/index.spec.js
```diff
@@ -7,52 +7,52 @@ const ipfsApi = require('ipfs-api')
...
+//
+//       fs.writeFileSync(path.join(npm2Path, 'ipfs'))
+//       delete require.cache[require.resolve('../src/node.js')]
+//       Node = require('../src/node.js')
+//       var node = new Node()
+//       assert.equal(node.exec, '/tmp/ipfsd-ctl-test/node_modules/ipfsd-ctl/node_modules/go-ipfs-dep/go-ipfs/ipfs')
+//       rimraf('/tmp/ipfsd-ctl-test', done)
+//     })
+//   })
+// })
```
diasdavid: Does bin-wrapper make it work flawlessly with npm2 and 3?

src/node.js
```diff
@@ -10,187 +10,218 @@ const shutdown = require('shutdown')
...
+  fs.readFile(join(path, 'config'), (err, data) => {
+    if (err) return done(err)
+    // Consistent error handling
+    try {
+      const parsed = JSON.parse(data)
+      return done(null, parsed)
+    } catch (err) {
+      return done(err)
+    }
+  })
```
diasdavid: why move from sync to async in this one?

src/node.js
```diff
@@ -10,187 +10,218 @@ const shutdown = require('shutdown')
...
-    this.opts = opts || {}
-    this.exec = process.env.IPFS_EXEC || ipfsDefaultPath
-    this.subprocess = null
-    this.initialized = fs.existsSync(path)
-    this.clean = true
-    this.env = Object.assign({}, process.env, {IPFS_PATH: path})
-
-    if (this.opts.env) Object.assign(this.env, this.opts.env)
-  }
+function Node (path, opts, disposable) {
```
dignifiedquire: let's keep it a `class` please

package.json
JGAntunes: Yeah, because despite having a run method (which is poorly named in my opinion) this will only act like a search for the binary and run a test command to check everything is ok (usually `--version` or `version` in our case), not considering any outputs and such. So we still need something to run the command itself and, as we already use `subcomandante`, I assume it's probably the best choice.

src/node.js
JGAntunes: This should probably be removed. `bin-wrapper` has support to check the binary version against what you're expecting, but unfortunately it doesn't allow us to set the arguments used to run the version command and it is currently defaulting to `--version`.

test/index.spec.js
JGAntunes: Sure thing :+1:

test/index.spec.js
JGAntunes: Yes because it's no longer up to npm where the ipfs binary is stored. Currently, unless you provide an IPFS_EXEC path, the path for binary defaults to `./vendor` under the module root directory. Let me know if you would prefer something else.

src/node.js
JGAntunes: While I was debugging why 0.4.2 didn't work ended up changing this and didn't change it back. But any particular reason why was this sync?

src/node.js
JGAntunes: Will do, sorry :+1:

JGAntunes:
@diasdavid as for the documentation, sure! I'll give it a look once all this stuff is sorted out.

src/node.js
diasdavid: Just avoiding callbacks when there is no need :)

test/index.spec.js
diasdavid: 👍

diasdavid:
Thank you :)

config.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
```
dignifiedquire: All files should be under `src`

JGAntunes:
Ok, let me know what you all think. Still missing documentation and further testing @diasdavid @dignifiedquire

JGAntunes:
While building some more tests I've noticed there's an issue with the way `subcomandante` handles the `close` event.
https://github.com/jbenet/node-subcomandante/blob/master/fork.js#L19
It may emit a new error when no further events are expected. This makes it practically impossible to handle it correctly as it is emitted after the `close` and `end` events and one doesn't know what to expect next.

diasdavid:
@JGAntunes Do we still need to use subcomandante with bin-wrapper?

Let's make this PR work with 0.4.3-rc1, so that we can use it to test js-ipfs-api as well

Do you need help with the documentation part?

JGAntunes:
@diasdavid we still need something to run and exec the commands, the bin-wrapper `run` method only serves for the single purpose of testing the binary and not much more. Although, if we still want to use subcomandante, we need to fix the way the exit code is handled.

Alright, I'll update the binary for 0.4.3-rc1

I was waiting for some feedback on the matters above so I can start documenting stuff. If the API will remain unchanged I can start working on that right away.

src/config.js
```diff
@@ -0,0 +1,16 @@
...
+const config = {
+  version: 'v0.4.1', // ipfs dist version
+  defaultExecPath: path.join(__dirname, '..', 'vendor'), // default path for the downloaded exec
+  gracePeriod: 7500 // amount of ms to wait before sigkill
+}
+
+config.baseUrl = 'https://dist.ipfs.io/go-ipfs/' +
+  config.version +
+  '/go-ipfs_' +
+  config.version + '_'
```
dignifiedquire: Could use `path.join`

```js
const distPage = 'https://dist.ipfs.io'
config.baseUrl = path.join(distPage, 'go-ipfs', config.version, `go-ipfs_${config.version}`)
```

src/node.js
```diff
@@ -56,31 +48,63 @@ module.exports = class Node {
...
+      if (err) {
+        // If no done callback return error to the error listener
+        if (!done) return listeners.error(err)
+        return done(err)
+      }
+      const ps = run(this.exec, args, opts)
+      const listenerKeys = Object.keys(listeners)
+      listenerKeys.forEach((key) => ps.on(key, listeners[key]))
+      // If done callback return process
+      if (done) return done(null, ps)
```
dignifiedquire: no need for `return`?

diasdavid:
@JGAntunes you can run any command with bin-wrapper -- https://www.npmjs.com/package/bin-wrapper#runcmd-callback -- too, running run('daemon', () =>) will keep the daemon process till it is closed.  Check how Akasha did it https://github.com/AkashaProject/ipfs-connector/blob/master/src/IpfsConnector.js

dignifiedquire:
@diasdavid we just need to ensure that it closes the process when the parent process dies (which is why node-subcommandante exists in the first place)

JGAntunes:
@diasdavid if you check the bin-wrapper source code the `run` method will only run as a binary check, it is not appropriate to run complex commands such as those needed here https://github.com/kevva/bin-wrapper/blob/master/index.js#L123. It will basically check the exit code of the command and that's it, nothing more.

I've already looked at how Akasha did it and if you look at it they also don't use bin-wrapper `run` to run anything at all besides the binary check - https://github.com/AkashaProject/ipfs-connector/blob/edd901e5b4bff9ed2783f44cb82797f49d68d54b/src/IpfsBin.js#L24. For the commands they still rely on a regular `childProcess.spawn` - https://github.com/AkashaProject/ipfs-connector/blob/master/src/IpfsConnector.js#L112

diasdavid:
@JGAntunes got it, thank you :)

diasdavid:
> I was waiting for some feedback on the matters above so I can start documenting stuff. If the API will remain unchanged I can start working on that right away.

We need to at least:
- spawn daemon - spawn a deamon with a repo from a 'path'
- spawn daemon and get a js-ipfs-api object that is connected to it
- spawn disposable daemon (optional config) - a daemon with a fresh repo
- spawn disposable daemon and get a js-ipfs-api object that is connected to it

haadcode:
We also need:
- make sure the daemon process gets killed in all cases
- daemon respects IPFS_PATH env variable (ie. spawn a daemon using repo from IPGS_PATH)

src/config.js
JGAntunes: Unfortunately `path.join` when used on URLs strips down the `//` to `/` and invalidates it. The result would be `https:/dist.ipfs.io/go-ipfs/v0.4.1/go-ipfs_v0.4.1`. I can leave the protocol aside, join it later and join everything else with `path.join` if you think that's worth it.

dignifiedquire: oh right, I forgot about that :( No it's not worth it then, but you could instead do `[..].join('/')`

JGAntunes: Yeah, true! I was also thinking on using template strings, something like:
```
const version = config.version
config.baseUrl = `https://dist.ipfs.io/go-ipfs/${version}/go-ipfs_${version}_`
```
Do you prefer this or the `Array.join`?

dignifiedquire: I think the just template strings is actually much more readable in this case :)

JGAntunes:
This PR should fix the issues around node-subcomandate https://github.com/jbenet/node-subcomandante/pull/5. FYI @dignifiedquire @diasdavid

package.json
dignifiedquire: I just filed an issue with node-core about not cleaning up spawned processes. So maybe in a near future we don't need it anymore :) 
https://github.com/nodejs/node/issues/7951

dignifiedquire:
@JGAntunes can you please add the `aegir@6` update to this PR

dignifiedquire:
@JGAntunes as soon as https://github.com/sindresorhus/execa/pull/53 is merged and released I suggest using that instead of subcomandante to spawn and control the daemon. It's a much nicer and well maintained module.

JGAntunes:
Hey folks, sorry for being kind of away towards the end of this week. @dignifiedquire will update the aegir module during this weekend and make the change to execa once your PR is merged :+1: I'll also try to give it a go on the documentation.

diasdavid:
feature request:

- [ ] add the binary path on package json ("bin":<path>) so that a user can npm i ipfsd-ctl and get the go-ipfs on their PATH

JGAntunes:
@diasdavid sounds like a cool feature but I got some concerns about it. Specifically, right now, the binary is only downloaded when it is needed (when the first command runs through the ipfsd-ctl) since this is handled by the `bin-wrapper`, don't know if we can change this or not.

JGAntunes:
@dignifiedquire @diasdavid bumped the aegir version and made the change for `execa` and, for now, it looks good!  The api used here - https://github.com/JGAntunes/js-ipfsd-ctl/blob/75f8d6910593e937733dcecbb81283aff1874ab0/src/exec.js - is still a bit strange and needs some improvements which I'll work on tomorrow. I also have to solve the new linting issues that appeared on the test file.

Meanwhile, since we're bringing some breaking changes, by looking at the API, should we consider changing `replaceConf` to `replaceConfig` on the node? (for the sake of consistency since we've got `setConfig` and `getConfig`).

jbenet:
Btw, the bin should probably be downloaded from a hardcoded dist.ipfs.io
hash, that way we know it's version locked and safe (not sure if this
happens now)

diasdavid:
@JGAntunes the replaceConf to replaceConfig sounds good, but better is to write on the readme all the function signatures that it exposes (or should expose), so that the 🚲🏠 can be efficient.

@jbenet yep, that is what happens :)

JGAntunes:
@diasdavid check the last commit yesterday, started working on the docs and that was the first thing I did ;)

README.md
```diff
@@ -36,17 +38,53 @@ IPFS daemons are already easy to start and stop, but this module is here to do i
...
+
+#### ipfs.daemonPid()
+
+#### ipfs.getConfig(key, done)
+
+#### ipfs.setConfig(key, value, done)
+
+#### ipfs.replaceConf(file, done)
+
+#### ipfs.version(done)
```
diasdavid: @dignifiedquire @haadcode please review the above, for methods you would like to see on ipfsd-ctl :)

haadcode: To be honest, I would like to use:
`ipfsd.start()`
`ipfs.stop()`

and that's it. start() would automatically init a repo if one doesn't exist at the given path.

However, I reckon that this would require a bigger re-think of the library and as I don't want to bikeshed it too much, I'm happy to use the current API if that's an easier way to bring this to a working state.

diasdavid: @haadcode now it is the time to bikeshed, since we have the opportunity to make a new version of ipfsd-ctl. Changing API is always harder as adoption increases, because that means breaking the interface for more people and harder to migrate. You are one of the primary users of this module and so, your input is super valuable when it comes to its API.

README.md
```diff
@@ -36,17 +38,53 @@ IPFS daemons are already easy to start and stop, but this module is here to do i
...
+
+## ipfsd
+
+#### ipfsd.version(done)
+
+#### ipfsd.local(path, done)
+
+#### ipfsd.disposableApi(opts, done)
+
+#### ipfsd.disposable(opts, done)
```
diasdavid: Let's have a:

```js
const node = ipfsd.create(options)
```

And set defaults for it to be a disposable daemon (random ports, random id, repo on the tmp folder and so on0

README.md
```diff
@@ -36,17 +38,53 @@ IPFS daemons are already easy to start and stop, but this module is here to do i
...
+
+## API
+
+## ipfsd
+
+#### ipfsd.version(done)
+
+#### ipfsd.local(path, done)
+
+#### ipfsd.disposableApi(opts, done)
```
diasdavid: instead of this. Let's use the call below (https://github.com/ipfs/js-ipfsd-ctl/pull/89/files#r75596364) and then do:

```
const ctl = node.apiCtl() 
// returns an ipfs-api instance lined to the daemon
```

README.md
```diff
@@ -36,17 +38,53 @@ IPFS daemons are already easy to start and stop, but this module is here to do i
...
+
+If you want to use an existing ipfs installation you can set `$IPFS_EXEC=/path/to/ipfs` to ensure it uses that.
+
+## API
+
+## ipfsd
+
+#### ipfsd.version(done)
+
+#### ipfsd.local(path, done)
```
diasdavid: These should be methods of the [`node` object](https://github.com/ipfs/js-ipfsd-ctl/pull/89/files#r75596364)

README.md
```diff
@@ -36,17 +38,53 @@ IPFS daemons are already easy to start and stop, but this module is here to do i
...
+
+#### ipfs.init(initOpts, done)
+
+#### ipfs.shutdown(done)
+
+#### ipfs.startDaemon(done)
+
+#### ipfs.stopDaemon(done)
+
+#### ipfs.daemonPid()
```
diasdavid: Does it have to be a func?

README.md
```diff
@@ -36,17 +38,53 @@ IPFS daemons are already easy to start and stop, but this module is here to do i
...
+
+#### ipfsd.version(done)
+
+#### ipfsd.local(path, done)
+
+#### ipfsd.disposableApi(opts, done)
+
+#### ipfsd.disposable(opts, done)
+
+## Ipfs(path, opts, disposable)
```
diasdavid: `s/Ipfs/IPFSNode` (being verbose in some cases pays off because we have a constant name overload)

README.md
```diff
@@ -36,17 +38,53 @@ IPFS daemons are already easy to start and stop, but this module is here to do i
...
+
+#### ipfs.init(initOpts, done)
+
+#### ipfs.shutdown(done)
```
diasdavid: what is the different between shutdown and stopDaemon?

README.md
```diff
@@ -36,17 +38,53 @@ IPFS daemons are already easy to start and stop, but this module is here to do i
...
+
+#### ipfs.daemonPid()
+
+#### ipfs.getConfig(key, done)
+
+#### ipfs.setConfig(key, value, done)
+
+#### ipfs.replaceConf(file, done)
+
+#### ipfs.version(done)
```
diasdavid: needs a method to wipeout the repo

diasdavid:
@JGAntunes Reviewed the API. Could you go through the https://github.com/ipfs/community/blob/master/js-commit-guidelines.md and reword your commits? thank you :)

README.md
diasdavid: also, the `done` should be the callback signature

README.md
JGAntunes: @diasdavid it can be a prop if you think it's more suited

README.md
JGAntunes: @diasdavid looking at it I think `shutdown` shouldn't be listed in the public IpfsNode API, as it is the handler to clean after a disposable node is killed. Going to remove it from the docs and probably rename de method accordingly.

JGAntunes:
@diasdavid sorry, didn't know about it. Should every commit be renamed? Isn't this going to be squashed into a single one?

README.md
JGAntunes: The version and local methods? @diasdavid

dignifiedquire:
> @diasdavid sorry, didn't know about it. Should every commit be renamed? Isn't this going to be squashed into a single one?

I think it's fine to leave the commits for now as they are, we can squash them into a couple of nice commits when all the work is done.

JGAntunes:
Sorry everyone, been away for the past two weeks. Hopefully I can close this on the following days.

JGAntunes:
Ok, I've just pushed the latest changes, let me know what you think @diasdavid @dignifiedquire @haadcode. Added the postinstall hook, so this way the binary will be downloaded right after this module is installed. I've got a couple of questions though, mainly to @diasdavid:

* https://github.com/ipfs/js-ipfsd-ctl/pull/89#discussion_r75596364 - We replaced the `disposable` methods from the daemon with the `create` one. Should we still maintain the `ipfsd.local(options)`? If the answer is no, then what should be the default behaviour for the `ipfsd.create()` method? Because right now it's creating a disposable daemon but imagine you want a local one using what you've defined through the `IPFS_PATH`.
* https://github.com/ipfs/js-ipfsd-ctl/pull/89#discussion_r75596413 - Changed the names of the classes on the `README`, let me know what you think.
* https://github.com/ipfs/js-ipfsd-ctl/pull/89#discussion_r75596418 - Probably a stupid question, but something that will `rm -rf` the repo path right?
* https://github.com/ipfs/js-ipfsd-ctl/pull/89#issuecomment-239709652- This feature request may be handled now, there are a couple of issues though. First we've got the directory into which the binary is downloaded, which is in the config file, but we can hardcode it if you prefer. Also, we've got the possibility that the user has the IPFS_EXEC env var defined, which will lead to no binary being downloaded, hence, no path to use on the `bin` field.

Finally, a couple of important stuff before this is released:
* We should probably wait until IPFS v0.4.3 is released since I noticed that there's only one release candidate available each time and each time there's a bump the whole module breaks.
* The `execa` dep with the clean feature wasn't released yet, so we're currently depending on the master branch of the repo which may be a problem.

Sorry writing so much btw :smile:

dignifiedquire:
Thank you @JGAntunes for the great work here. We have updated [go-ipfs-dep](https://github.com/ipfs/npm-go-ipfs-dep) to include a lot of the things we learned from here and more, so we don't need to implement all the binary handling here :)
The internal improvements of using `execa` and your `exec.js` I have pulled in manually in https://github.com/ipfs/js-ipfsd-ctl/pull/140/commits/7b0952956015ffe2629bd283630924971bdfcfaa 

Again thank you for all the work and keeping with us. 

In terms of the actual interface changes I think we will want to go with the ideas here: https://github.com/ipfs/js-ipfs/issues/556
as we are attempting to unify the interfaces between spawning a js and a go daemon.

diasdavid:
Merged https://github.com/ipfs/js-ipfsd-ctl/pull/140, closing this one. @JGAntunes again thank you for helping out with these issues and my apologies for the huge lag from PR to actually getting something merged in, this module is highly sensible as it touches running processes at the OS level and so improvements have to happen slowly (execa vs subcomandante), as for the remaining features added, as @dignifiedquire noted, they were added to npm-go-ipfs-dep, which is now capable of picking the right dist through a package.json value.

Again, thank you <3

JGAntunes:
Np folks, glad I could help out anyway :+1:

PR #107, opened 2016-09-16
bugfixes and improvements (ipfs daemon flags)

jbenet:
d74045f (jbenet, 2 minutes ago)
    feat(startDeamon): allow passing flags to ipfs daemon

 9c4dbce (jbenet, 4 minutes ago)
    fix(startDaemon): fix the behavior of startDeamon

    - now waits for "Daemon is ready"
    - also tracks Gateway address
    - checks failures on startup, and returns an error if so
     - this was a serious bug-- would not detect errors.
    - fails correctly, without leaving the process hanging

 b523b1c (jbenet, 7 minutes ago)
    fix: rm unused var (thanks, linter)

 efa2aa4 (jbenet, 10 minutes ago)
    note(shutdown): comment on problematic func name

 d6ad480 (jbenet, 14 minutes ago)
    fix(shutdown): fixed bugs in stopDaemon

    - bug1: local var for the closure (this.subprocess set to null)
    - bug2: order problem: should set .on('close') handler before .kill()

 3c50894 (jbenet, 17 minutes ago)
    fix(tests): don't force changing test when src changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
   "license": "MIT",
   "dependencies": {
     "go-ipfs-dep": "0.4.1",
-    "ipfs-api": "^4.1.0",
+    "ipfs-api": "^9.0.0",
     "multiaddr": "^2.0.0",
     "rimraf": "^2.4.5",
     "run-series": "^1.1.4",
--- a/src/node.js
+++ b/src/node.js
@@ -118,38 +118,86 @@ module.exports = class Node {
   }
 
   startDaemon (done) {
-    parseConfig(this.path, (err, conf) => {
+    const node = this
+    parseConfig(node.path, (err, conf) => {
       if (err) return done(err)
 
-      this.subprocess = run(this.exec, ['daemon'], {env: this.env})
-        .on('error', (err) => {
-          if (String(err).match('daemon is running')) {
-            // we're good
-            done(null, ipfs(conf.Addresses.API))
-          } else if (String(err).match('non-zero exit code')) {
-            // ignore when kill -9'd
-          } else {
-            done(err)
-          }
-        })
-        .on('data', (data) => {
-          const match = String(data).trim().match(/API server listening on (.*)/)
-          if (match) {
-            this.apiAddr = match[1]
-            const addr = multiaddr(this.apiAddr).nodeAddress()
-            const api = ipfs(this.apiAddr)
-            api.apiHost = addr.address
-            api.apiPort = addr.port
-            done(null, api)
-          }
-        })
+      let stdout = ""
+
+      // strategy:
+      // - run subprocess
+      // - listen for API addr on stdout (success)
+      // - or an early exit or error (failure)
+      node.subprocess = run(node.exec, ['daemon'], {env: node.env})
+      node.subprocess.on('error', onErr)
+        .on('data', onData)
+
+      // done2 is called to call done after removing the event listeners
+      function done2 (err, val) {
+        node.subprocess.removeListener('data', onData)
+        node.subprocess.removeListener('error', onErr)
+        if (err) {
+          node.killProcess(() => {}) // we failed. kill, just to be sure...
+        }
+        done(err, val)
+        done = () => {} // in case it gets called twice
+      }
+
+      function onErr (err) {
+        if (String(err).match('daemon is running')) {
+          // we're good
+          done2(null, ipfs(conf.Addresses.API))
+
+          // TODO: I don't think this case is OK at all...
+          // When does the daemon outout "daemon is running" ?? seems old.
+          // Someone should check on this... - @jbenet
+        } else if (String(err).match('non-zero exit code')) {
+          // exited with an error on startup, before we removed listeners
+          done2(err)
+        } else {
+          done2(err)
+        }
+      }
+
+      function onData (data) {
+        data = String(data)
+        stdout += data
+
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -44,7 +44,7 @@
   "license": "MIT",
   "dependencies": {
     "go-ipfs-dep": "0.4.1",
-    "ipfs-api": "^4.1.0",
+    "ipfs-api": "^9.0.0",
```
diasdavid: 🌟

src/node.js
```diff
@@ -109,6 +106,8 @@ module.exports = class Node {
   }
 
   // cleanup tmp files
+  // TODO: this is a bad name for a function. a user may call this expecting
+  // something similar to "stopDaemon()". consider changing it. - @jbenet
```
diasdavid: 👍

src/node.js
```diff
@@ -118,51 +117,106 @@ module.exports = class Node {
...
 
-  startDaemon (done) {
-    parseConfig(this.path, (err, conf) => {
+  startDaemon (flags, done) {
+    if (typeof flags === 'function' && typeof done === 'undefined') {
+      done = flags
+      flags = []
+    }
+
+    const node = this
```
diasdavid: Seems that having access to the `this` of this func is not necessary. Rather use `{}` to initialize it instead?

src/node.js
```diff
@@ -118,51 +117,106 @@ module.exports = class Node {
...
+        stdout += data
+
+        if (!data.trim().match(/Daemon is ready/)) {
+          return // not ready yet, keep waiting.
+        }
+
+        const apiM = stdout.match(/API server listening on (.*)\n/)
+        if (apiM) {
+          // found the API server listening. extract the addr.
+          node.apiAddr = apiM[1]
```
diasdavid: This is a bit ninja. the daemon leaves a 'api' file on the repo with the multiaddr that is listening. It would be more safe to use that than trusting that this string will be always the same.

src/node.js
```diff
@@ -118,51 +117,106 @@ module.exports = class Node {
...
+        if (gatewayM) {
+          // found the Gateway server listening. extract the addr.
+          node.gatewayAddr = gatewayM[1]
+        }
+
+        const addr = multiaddr(node.apiAddr).nodeAddress()
+        const api = ipfs(node.apiAddr)
+        api.apiHost = addr.address
+        api.apiPort = addr.port
+        done2(null, api)
```
diasdavid: The js-ipfs-api is able to dial to multiaddrs, no need for parsing. It is a good practice to keep using multiaddrs for addrs everywhere

src/node.js
```diff
@@ -118,51 +117,106 @@ module.exports = class Node {
...
+
+      // done2 is called to call done after removing the event listeners
+      function done2 (err, val) {
+        node.subprocess.removeListener('data', onData)
+        node.subprocess.removeListener('error', onErr)
+        if (err) {
+          node.killProcess(() => {}) // we failed. kill, just to be sure...
+        }
+        done(err, val)
+        done = () => {} // in case it gets called twice
```
diasdavid: If we get multiple errors (queued) do we want to do node.killProcess several times? What about having a flag 'isDead'?

src/node.js
```diff
@@ -118,51 +117,106 @@ module.exports = class Node {
     }
   }
 
-  startDaemon (done) {
-    parseConfig(this.path, (err, conf) => {
+  startDaemon (flags, done) {
```
diasdavid: callbacks at the top level of the func should be `callback` just to keep being consistent with the rest of js ecosystem (I do notice that applies to other parts of this module, it just means it was never refactored)

src/node.js
```diff
@@ -118,51 +117,106 @@ module.exports = class Node {
...
+      // strategy:
+      // - run subprocess
+      // - listen for API addr on stdout (success)
+      // - or an early exit or error (failure)
+      node.subprocess = run(node.exec, args, {env: node.env})
+      node.subprocess.on('error', onErr)
+        .on('data', onData)
+
+      // done2 is called to call done after removing the event listeners
+      function done2 (err, val) {
```
diasdavid: This doesn't seem right. This function gets called on a successful start of a node or a on insuccessful one. Can we have instead a 'failure' func and a 'success' func instead?

src/node.js
jbenet: this was the way it was. i did not change it

src/node.js
jbenet: this would change the expected api-- i dont necessarily want to break things expecting this to be here

src/node.js
jbenet: okay sounds good

src/node.js
jbenet: yeah i dont want to change this across modules

src/node.js
diasdavid: ACK. Opened a issue so that it doesn't get lost - https://github.com/ipfs/js-ipfsd-ctl/issues/108

src/node.js
diasdavid: 👍

src/node.js
jbenet: in both cases the same thing -- removing listeners -- has to happen, you would have two functions with the same code. this changes from the old behavior (which was just to call `done`, to this wrapper on `done` (`done2`, not the best name))

PR #110, opened 2016-09-16
Add gateway address to the api instance

haadcode:
This PR will add the gateway address to the daemon instance. This is useful when the gateway is bound to `.../tcp/0` (random port) to know later where the gateway is located.


haadcode:
This was already added by @jbenet in 0.15.0

PR #124, opened 2016-11-14
Fail gracefully when config file is not found

scott113341:
In the case that no config file exists, the error generated by `fs.readFileSync` is uncaught.  This results in the `done` callback not being called.


diasdavid:
thank you @scott113341 :)

PR #139, opened 2016-12-20
Call 'close' only once when the process is killed

haadcode:
Call close only was in order to avoid double callback when the process is killed.


dignifiedquire:
@haadcode any idea why ci is unhappy?

dignifiedquire:
LGTM, CI will be happy once #140 is merged

dignifiedquire:
I pulled this into #140

Issue #95, opened 2016-08-15
Enumerate the tests in order to make sure that execa is as robust as subcommandante


diasdavid:
As discussed in August 15 js-ipfs sprint meeting. @jbenet please enumerate all the tests you would like to see for: https://github.com/ipfs/js-ipfsd-ctl/pull/89

//cc @jgantunes

jbenet:
@diasdavid no sorry I don't have time for this in the next 2 weeks. If you
want me to this it probably won't get done for a while.

jbenet:
The burden of proof also lies on whoever wants to _change_ the behavior.
Its up to that PR to ensure correctness

diasdavid:
Ok, I'll handle this myself with @jgantunes

dignifiedquire:
I found out the hard way that it isn't :(

diasdavid:
Could you elaborate?

diasdavid:
Nvm, saw the commits on https://github.com/ipfs/js-ipfsd-ctl/pull/140#issuecomment-268774809, closing this one then :)

PR #140, opened 2016-12-20
feat: better stream handling

dignifiedquire:
- upgrade aegir
- upgrade dependencies
- handle stream errors when setting the config
- use pump instead of pipe

I've also pulled in the changes from subcomandante to execa from #89

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,13 +45,13 @@
   "license": "MIT",
   "dependencies": {
     "async": "^2.1.4",
-    "execa": "^0.5.0",
     "go-ipfs-dep": "0.4.4",
     "ipfs-api": "^12.0.3",
     "multiaddr": "^2.1.0",
     "once": "^1.4.0",
     "rimraf": "^2.5.4",
-    "shutdown": "^0.2.4"
+    "shutdown": "^0.2.4",
+    "subcomandante": "^1.0.5"
   },
   "devDependencies": {
     "aegir": "^9.3.0",
--- a/package.json
+++ b/package.json
@@ -46,8 +46,8 @@
   "dependencies": {
     "async": "^2.1.4",
     "go-ipfs-dep": "0.4.4",
-    "ipfs-api": "^12.0.3",
-    "multiaddr": "^2.1.0",
+    "ipfs-api": "^12.1.2",
+    "multiaddr": "^2.1.1",
     "once": "^1.4.0",
     "rimraf": "^2.5.4",
     "shutdown": "^0.2.4",
@@ -59,7 +59,7 @@
     "is-running": "^2.1.0",
     "mkdirp": "^0.5.1",
     "multihashes": "^0.3.1",
-    "pre-commit": "^1.2.1"
+    "pre-commit": "^1.2.2"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
   "devDependencies": {
     "aegir": "^9.3.0",
     "chai": "^3.5.0",
+    "is-running": "^2.1.0",
     "mkdirp": "^0.5.1",
     "multihashes": "^0.3.1",
     "pre-commit": "^1.2.1"
--- a/package.json
+++ b/package.json
@@ -7,9 +7,10 @@
     "lint": "aegir-lint",
     "coverage": "aegir-coverage",
     "test": "aegir-test --env node",
-    "release": "aegir-release --env node",
-    "release-minor": "aegir-release --type minor --env node",
-    "release-major": "aegir-release --type major --env node",
+    "docs": "aegir-docs",
+    "release": "aegir-release --env node --docs",
+    "release-minor": "aegir-release --type minor --env node --docs",
+    "release-major": "aegir-release --type major --env node --docs",
     "coverage-publish": "aegir-coverage publish"
   },
   "engines": {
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -45,19 +43,22 @@
...
-    "run-series": "^1.1.4",
     "shutdown": "^0.2.4",
     "subcomandante": "^1.0.5"
   },
   "devDependencies": {
-    "aegir": "^9.2.1",
+    "aegir": "^9.3.0",
+    "chai": "^3.5.0",
     "mkdirp": "^0.5.1",
+    "multihashes": "^0.3.1",
```
haadcode: Is multihashes needed somewhere else? I don't see it being used in the tests.

haadcode:
One comment and once CI is happy, LGTM.

package.json
```diff
@@ -45,19 +43,20 @@
...
+    "async": "^2.1.4",
+    "execa": "^0.5.0",
     "go-ipfs-dep": "0.4.4",
     "ipfs-api": "^12.0.3",
     "multiaddr": "^2.1.0",
     "once": "^1.4.0",
     "rimraf": "^2.5.4",
-    "run-series": "^1.1.4",
-    "shutdown": "^0.2.4",
-    "subcomandante": "^1.0.5"
```
diasdavid: @jbenet invested a lot of time making sure we had a thing that would not leave zombie IPFS processes in user machines. It was what blocked https://github.com/ipfs/js-ipfsd-ctl/pull/89 all this time.

Until we have comprehensive tests to make sure that you can spawn a bunch of daemons and make sure they don't become zombies, we can't move to execa. Historically, a way to test this was with `station`

If I understand correctly, migrating to `execa` is not a requirement for any particular feature.

diasdavid: As for the tests: https://github.com/ipfs/js-ipfsd-ctl/issues/95

src/index.js
```diff
@@ -5,6 +5,14 @@ const join = require('path').join
...
+  'Addresses.Swarm': ['/ip4/0.0.0.0/tcp/0'],
+  'Addresses.Gateway': '',
+  'Addresses.API': '/ip4/127.0.0.1/tcp/0',
+  disposable: true,
+  init: true
+}
+
 function tempDir () {
   return join(os.tmpdir(), `ipfs_${String(Math.random()).substr(2)}`)
 }
```
diasdavid: Is the path to tmpDir available through some call?

src/node.js
```diff
@@ -64,25 +83,20 @@ module.exports = class Node {
     this.initialized = fs.existsSync(path)
     this.clean = true
     this.env = Object.assign({}, process.env, {IPFS_PATH: path})
+    this.disposable = disposable
```
diasdavid: add to the list of props `path` as the path used for the repo

src/node.js
```diff
@@ -93,148 +107,108 @@ module.exports = class Node {
...
   // cleanup tmp files
-  // TODO: this is a bad name for a function. a user may call this expecting
-  // something similar to "stopDaemon()". consider changing it. - @jbenet
-  shutdown (done) {
-    if (!this.clean && this.disposable) {
-      rimraf(this.path, (err) => {
-        if (err) throw err
-        done()
-      })
+  shutdown (callback) {
```
diasdavid: still a `bad name`, check the deleted comment

diasdavid: but independent of the name, it should be document of what it is doing

src/node.js
dignifiedquire: not sure, I didn't change anything from that logic

src/node.js
dignifiedquire: I disagree I think it's actuallly a pretty fitting name, in terms of docs nothing changed

package.json
dignifiedquire: I don't see how the burden of proof lies on the one changing it, as I haven't seen any proof subcomandante being as good as you say it is. There are tests for the cleanup behavior inside execa, which I wrote when implementing. I also did manual tests with execa inside reusing the testsuite from subcomandante so those cases all work. 

So far all I heard was 'there were extensive tests' but I haven't seen or read what those actually were, so please tell which scenarios you are concerned about and I can add tests for them

src/node.js
dignifiedquire: this comment was meant for the one above

src/node.js
diasdavid: There are no docs (am I missing something?)

dignifiedquire: There is a comment above which kinda describes what it does, no other docs currently, correct

src/node.js
dignifiedquire: > add to the list of props path as the path used for the repo

already there if you look at the full file

dignifiedquire:
Added documentation https://ipfs.github.io/js-ipfsd-ctl/

package.json
dignifiedquire: @diasdavid I've ported the test suite from node-subcomandante, to run against the new `src/exec.js` to add a baseline of coverage to this module.

dignifiedquire:
Looks like I can't run the last test on travis: 

```
    ✓ SIGTERM kills hang (412ms)
trap: SIGHUP: bad trap
    1) SIGKILL kills survivor
```

package.json
dignifiedquire: Soo even though those tests pass fine, it seems thing are wonky right now. I have 5 ipfs instances running after a single `npm test` run :( 

I will revert the switch to execa tomorrow.

dignifiedquire:
@diasdavid this should be ready now, we are back to subcomandante and have more tests and docs

Issue #143, opened 2017-01-03
Doesn't kill daemon / sub-process


haadcode:
Previously, js-ipfsd-ctl killed all sub-processes upon process exit. However, with the recent refactoring, that doesn't seem to be the case anymore :/

If you run the tests, for example only the 'start daemon' ones (describe.only('startDaemon')), the tests will pass but if you then do `ps -A|grep ipfs` you will see that there are still ipfs processes running (3 if I counted correctly). I'm on OSX, running js-ipfsd-ctl from master.

This is a **major** issue.

@jbenet kept trying to say that this is exactly what we need to avoid, because the old code was crafted over time to handle this exact case. We need to put fixing this on highest priority because it basically effects every program that uses js-ipfsd-ctl, leaving processes hanging.

cc @dignifiedquire @diasdavid @VictorBjelkholm

dignifiedquire:
This is strange, because we didn't actually change the spawning, I reverted those changes and we still use `subcomandante` for the exiting.

haadcode:
Could we try to verify this on various platforms? Can you reproduce it locally? Can we reproduce it on CI?

diasdavid:
I'm seeing this error two, but here is further debugging:

- The test `SIGKILL kills survivor` just hangs, leaving a couple of ipfs processes running. 
- Skipping the previous tests makes everything pass
- Running the tests again will make `1) start and stop` fail with:

```sh
 1) daemons startDaemon start and stop:
     Error: Error: ipfs configuration file already exists!
Reinitializing would overwrite your keys.


      at Object.done (src/exec.js:22:27)
      at Stream.listeners.done.once (src/exec.js:44:18)
      at Stream.f (node_modules/once/once.js:25:25)
      at Socket.reemit (node_modules/duplexer/index.js:70:25)
      at Pipe._handle.close [as _onclose] (net.js:498:12)
```

diasdavid:
@haadcode it was `is-running` https://github.com/ipfs/js-ipfsd-ctl/pull/144

diasdavid:
fix on https://github.com/ipfs/js-ipfsd-ctl/pull/144

PR #144, opened 2017-01-04
Lingering daemons

diasdavid:
The bug reported by https://github.com/ipfs/js-ipfsd-ctl/issues/143 was introduced with https://github.com/ipfs/js-ipfsd-ctl/commit/a0adc8225164e4edfbe9346cc93c50c3f6bd71c2, I'm still super intrigued how the tests passed the scripts pre-release, but truth is that I'm not being able to get to pass now.

Found a issue with the `psExpect` function where the grace countdown had to be `--grace`, otherwise it would never reach 0.

The test failing is that SIGKILL is not effectively killing the process, however the test here is straight out copy of https://github.com/jbenet/node-subcomandante/blob/master/test/test.js#L73-L97 and I've been tinkering back and forward, making both match and I can't seem to figure out exactly where the issue is coming from.

Going to visit this back tomorrow, otherwise we will revert the commit and release a working version.

### DIFF
--- a/test/exec.spec.js
+++ b/test/exec.spec.js
@@ -9,35 +9,33 @@ const path = require('path')
 const exec = require('../src/exec')
 
 const survivor = path.join(__dirname, 'survivor')
-const hang = 'tail -f /dev/null'.split(' ')
 
 function token () {
   return Math.random().toString().substr(2)
 }
 
-function psExpect (pid, expect, grace, cb) {
+function psExpect (pid, expect, grace, callback) {
   setTimeout(() => {
     const actual = isrunning(pid)
 
     if (actual !== expect && grace > 0) {
-      psExpect(pid, expect, grace--, cb)
-      return
+      return psExpect(pid, expect, --grace, callback)
     }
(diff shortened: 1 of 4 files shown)


test/exec.spec.js
```diff
@@ -9,23 +9,21 @@ const path = require('path')
...
 
-function psExpect (pid, expect, grace, cb) {
+function psExpect (pid, expect, grace, callback) {
   setTimeout(() => {
     const actual = isrunning(pid)
 
     if (actual !== expect && grace > 0) {
-      psExpect(pid, expect, grace--, cb)
-      return
+      return psExpect(pid, expect, --grace, callback)
```
diasdavid: This was never finishing as grace was never decremented for the next cycles

dignifiedquire:
Can you please remove all the changes that are not actually relevant for this fix?

diasdavid:
@dignifiedquire removed the unnecessary console.logs and the named callback. All the other stuff are good, no?

PR #145, opened 2017-01-04
Add gateway address to the returned IPFS api

haadcode:
This PR will add a properties to get the address the gateway was started at and can be accessed with:

```javascript
ipfs.gatewayHost
ipfs.gatewayPort
```

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -198,6 +198,7 @@ class Node {
         },
         data: (data) => {
           const match = String(data).trim().match(/API server listening on (.*)/)
+          const gwmatch = String(data).trim().match(/Gateway (.*) listening on (.*)/)
 
           if (match) {
             this.apiAddr = match[1]
@@ -205,6 +206,13 @@ class Node {
             this.api = ipfs(this.apiAddr)
             this.api.apiHost = addr.address
             this.api.apiPort = addr.port
+
+            if (gwmatch) {
+              this.gatewayAddr = gwmatch[2]
+              const addr = multiaddr(this.gatewayAddr).nodeAddress()
+              this.api.gatewayHost = addr.address
+              this.api.gatewayPort = addr.port
+            }
 
             callback(null, this.api)
           }
--- a/src/node.js
+++ b/src/node.js
@@ -197,8 +197,9 @@ class Node {
           }
         },
         data: (data) => {
-          const match = String(data).trim().match(/API server listening on (.*)/)
-          const gwmatch = String(data).trim().match(/Gateway (.*) listening on (.*)/)
+          const str = String(data).trim()
+          const match = str.match(/API server listening on (.*)/)
+          const gwmatch = str.match(/Gateway (.*) listening on (.*)/)
 
           if (match) {
             this.apiAddr = match[1]
--- a/src/node.js
+++ b/src/node.js
@@ -87,10 +87,30 @@ class Node {
     this.clean = true
     this.env = Object.assign({}, process.env, {IPFS_PATH: path})
     this.disposable = disposable
+    this._apiAddr = null
+    this._gatewayAddr = null
 
     if (this.opts.env) {
       Object.assign(this.env, this.opts.env)
     }
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


dignifiedquire:
CI is very unhappy here

haadcode:
Fixed!

src/node.js
```diff
@@ -198,6 +198,7 @@ class Node {
         },
         data: (data) => {
           const match = String(data).trim().match(/API server listening on (.*)/)
+          const gwmatch = String(data).trim().match(/Gateway (.*) listening on (.*)/)
```
dignifiedquire: Lets not duplicate work. The trim and string cast are happening twice now.

src/node.js
```diff
@@ -206,6 +207,13 @@ class Node {
             this.api.apiHost = addr.address
             this.api.apiPort = addr.port
 
+            if (gwmatch) {
+              this.gatewayAddr = gwmatch[2]
+              const addr = multiaddr(this.gatewayAddr).nodeAddress()
```
dignifiedquire: How about only exposing `this.gatewayAddr = multiaddr(gwmatch[2])`. That way we have the standard multiaddr we are using everywhere and you can pull the details you need at any point.

src/node.js
haadcode: Will fix this.

src/node.js
haadcode: I believe following the current convention as per api host and port is a better approach for consistency.

We can do what you propose in separate PR, bearing in mind it would be a public API change.

dignifiedquire: how is the a breaking change, currently the `gatewayAddr` doesn't exist from what I can tell?

haadcode: Jumped some steps here :) Let me try again:

We currently have (with this PR):
```javascript
ipfs.apiHost
ipfs.apiPort
ipfs.gatewayHost
ipfs.gatewayPort
```

If we would expose the addresses as multiaddr, we should make sure the properties are consistent:
```javascript
ipfs.apiAddr
ipfs.gatewayAddr
```

So with that, I would get rid of `ipfs.apiHost` and `ipfs.apiPort`, and that would be the breaking change.

Does that make sense? We could also add both `ipfs.gatewayAddr` and `ipfs.apiAddr` in this PR, but then we'd have the `apiHost` and `apiPort` still there and I think that'd be slightly more confusion to have same thing as multiple properties.

Let me know what you think!

dignifiedquire: How about exposing `gatewayAddr` and `apiAddr` as the main thing, and making the others simple helper functions which just depend on those?

haadcode: Changed the code now to expose `apiAddr` and `gatewayAddr` in the `Node` and to keep api compatibility, keep the separate variables for host and port in the `ipfs` instance. I'm not a huge fan of the monkey-patching solution here as it changed the public API of js-ipfs-api, but that's how it's always been, so I don't want to break that now.

@dignifiedquire take a look at let me know if you think this is good to go.

haadcode:
As noticed on IRC, this will change the public property `apiAddr` from string to multiaddr. We should prolly bump the version? Should I bump it to 0.19.0?

dignifiedquire:
version bump will happen on release.

diasdavid:
- [ ] Needs deps to be updated (npm-go-ipfs)

Is there PR to cope with the breaking change proposed here?

haadcode:
> Are there PRs to cope with the breaking change proposed here?

This will change the public API of `ipfsd-ctl`, not the ipfs API returned by it. Ie. nothing should break for js-ipfs-api. Do we test this module somewhere else than here in its own tests?

diasdavid:
Just to be sure, can you check if js-ipfs-api runs smoothly with these changes?

haadcode:
> Just to be sure, can you check if js-ipfs-api runs smoothly with these changes?

Sure thing! I'll add tests to js-ipfs-api for this today.

diasdavid:
https://github.com/ipfs/js-ipfs-api/pull/524/files#r101501630

PR #150, opened 2017-03-22
fix: Return node only after go-ipfs daemon is ready

haadcode:
Testing new ipfsd-ctl with Orbit I found a bug where the node is returned before the gateway address has been received and patched to the returned API. Ie. gatewayAddr was undefined.

This PR changes the callback logic so that we listen for "Daemon is ready" stdout from go-ipfs and only then return. Previously we returned when stdout matched "API listening...".

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -220,19 +220,22 @@ class Node {
           const str = String(data).trim()
           const match = str.match(/API server listening on (.*)/)
           const gwmatch = str.match(/Gateway (.*) listening on (.*)/)
+          const readyMatch = str.match(/Daemon is ready/)
 
           if (match) {
             this._apiAddr = multiaddr(match[1])
             this.api = ipfs(match[1])
             this.api.apiHost = this.apiAddr.nodeAddress().address
             this.api.apiPort = this.apiAddr.nodeAddress().port
+          }
 
-            if (gwmatch) {
-              this._gatewayAddr = multiaddr(gwmatch[2])
-              this.api.gatewayHost = this.gatewayAddr.nodeAddress().address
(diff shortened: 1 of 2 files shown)


src/node.js
```diff
@@ -220,19 +220,22 @@ class Node {
...
-            }
+          if (gwmatch) {
+            this._gatewayAddr = multiaddr(gwmatch[2])
+            this.api.gatewayHost = this.gatewayAddr.nodeAddress().address
+            this.api.gatewayPort = this.gatewayAddr.nodeAddress().port
+          }
 
+          if (readyMatch) {
             callback(null, this.api)
           }
```
diasdavid: `str.match` is non blocking, if you found that bug happening, it means that the output was captured before go-ipfs had written everything. I believe that this doesn't solve the issue (🕸🕷), only silences it by not calling the callback with an undefined parameter. Let's add a `else` to callback with a error signalling that the daemon was not found ready.

haadcode: Yeah, this part was getting called multiple times (twice in this case), first with "API listening..." and then with "Gateway (readonly) listening...\nDaemon is ready". Not sure why or if that's the intended behaviour, but adding an else and an error callback will just make it fail always (from user's perspective).

@dignifiedquire I believe you refactored the exec code last time, any ideas here?

diasdavid: You can use http://npmjs.org/once

diasdavid:
@haadcode just completed the task of slaying the very big dragon that was dormant in ipfsd-ctl. See how I solved it here: https://github.com/ipfs/js-ipfsd-ctl/pull/165. This was a fun one 🐲

haadcode:
Awesome! Thank you for slaying the 🐉 😄

PR #151, opened 2017-06-19
adds test getConfig without key, should get the whole config

thisconnect:
see https://ipfs.github.io/js-ipfsd-ctl/#nodegetconfig
If no key is passed, the whole config is returned as an object.

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -306,7 +306,7 @@ class Node {
   getConfig (key, callback) {
     if (typeof key === 'function') {
       callback = key
-      key = ''
+      key = 'show'
     }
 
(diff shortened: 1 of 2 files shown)


thisconnect:
getConfig with no key does currently not return the whole config, but `Error: Failed to get config value: " key has no attributes"`

thisconnect:
Expected this to pass, but it seems to fail on Node.js v6.x, but in a unrelated(?) test https://travis-ci.org/ipfs/js-ipfsd-ctl/jobs/244451299#L1622

Locally everything passes on Node.js v6.x :/

PR #154, opened 2017-06-22
Examples

thisconnect:


example.js
```diff
@@ -1,12 +1,12 @@
 'use strict'
```
diasdavid: Could you please move this example to the examples folder? I just realised now that it ended up in the root path

example.js
```diff
@@ -1,12 +1,12 @@
...
 
 // Start a disposable node, and get access to the api
-// print the node id, and kill the temporary daemon
+// print the node id
 
 // IPFS_PATH will point to /tmp/ipfs_***** and will be
 // cleaned up when the process exits.
 
-const ipfsd = require('ipfsd-ctl')
+const ipfsd = require('.')
```
diasdavid: Does this work? It should have to point to `src`

examples/id.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-var ipfsd = require('../index.js')
+var ipfsd = require('..')
```
diasdavid: same as above

examples/local.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-var ipfsd = require('../index.js')
+var ipfsd = require('..')
```
diasdavid: same as above

example.js
thisconnect: Yes this works, though in the Node.js docs `require('./')` with a slash is commonly used. As it is pointing to the root of this module Node.js will lookup the main field in package.json and so it finds `src/index.js`. 
https://nodejs.org/api/modules.html#modules_all_together

If the examples point to the root of the module, the src files and main field in package.json could be refactored and moved without changing the examples. But I'm happy to change the examples/* to point to `../src` if you want.

diasdavid:
Thank you @thisconnect :)

Issue #156, opened 2017-06-22
Error npm test fills tmp dir


thisconnect:
while running the tests over and over again the following error appears more often than it should (MacOS)

```
Uncaught AssertionError: expected [Error: Error: ipfs configuration file already exists!
Reinitializing would overwrite your keys.
```

some tests use a temporary dir 
```js
const dir = os.tmpdir() + `/tmp-${Math.ceil(Math.random() * 100)}`
```

This temporary dir gets quite full and seems to be the cause for the Error: ipfs configuration file already exists

```bash
ls /var/folders/vc/g0gp6cj15vg3v62bwyn3pdbc0000gp/T
ipfs_00923925270015502				tmp--100
ipfs_013408338790235286				tmp--12
ipfs_014171347120929045				tmp--14
ipfs_027350138789616674				tmp--16
ipfs_029120182442286202				tmp--18
ipfs_03184452858010567				tmp--20
ipfs_05955467903437994				tmp--22
ipfs_06095803926897725				tmp--23
ipfs_06740066115484877				tmp--24
ipfs_06795640497725586				tmp--27
ipfs_06993119814338011				tmp--28
ipfs_07298509408059695				tmp--31
ipfs_07456547692741511				tmp--33
ipfs_08634351266617024				tmp--35
ipfs_0888132984196397				tmp--43
ipfs_09132801630670873				tmp--44
ipfs_09625399003146295				tmp--55
ipfs_10890433268035649				tmp--56
ipfs_11631035864891426				tmp--64
ipfs_11924458870556665				tmp--65
ipfs_1258814504860306				tmp--66
ipfs_15681934416640764				tmp--70
ipfs_17131383315111948				tmp--72
ipfs_18804510593267376				tmp--75
ipfs_1948177500601116				tmp--77
ipfs_2147503549547347				tmp--78
ipfs_2232495460229984				tmp--83
ipfs_22882884304077367				tmp--87
ipfs_23387611613097037				tmp--88
ipfs_23837606108405018				tmp--9
ipfs_23894842144610107				tmp--94
ipfs_24736804932801193				tmp--99
ipfs_25719110414888413				tmp-15
ipfs_26157466717751476				tmp-17
ipfs_2641200796085763				tmp-19
ipfs_2773751683075154				tmp-20
ipfs_2803329334680442				tmp-21
ipfs_2936031290305703				tmp-26
ipfs_29968229753631714				tmp-29
ipfs_30119929804576495				tmp-3
[...]

PR #157, opened 2017-06-22
tests improves tmp dir

thisconnect:
This should fix https://github.com/ipfs/js-ipfsd-ctl/issues/156

This still does **NOT** fix the "occasional" random CI errors 
```
 1) daemons startDaemon starts the daemon and returns valid API and gateway addresses:
     Uncaught AssertionError: expected null to be an instance of Multiaddr
```
as seen here https://travis-ci.org/ipfs/js-ipfsd-ctl/builds/245233174 
or here https://travis-ci.org/ipfs/js-ipfsd-ctl/builds/244451297

### DIFF
--- a/test/index.spec.js
+++ b/test/index.spec.js
@@ -393,7 +393,7 @@ describe('daemons', () => {
 
   describe('startDaemon', () => {
     it('start and stop', (done) => {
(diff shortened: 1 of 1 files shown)


diasdavid:
Thank you @thisconnect, this reduces the chance of collision. Ideally we should just do better cleanup.

Issue #158, opened 2017-06-24
CI: random fail returns valid API and gateway addresses


thisconnect:
Ci failes "sometimes" on Node 4 sometimes on Node 6

See the 2 builds from another PR (https://github.com/ipfs/js-ipfsd-ctl/pull/153 should be unrelated to the test)
fails on Node 6 https://travis-ci.org/ipfs/js-ipfsd-ctl/builds/244451297
fails on Node 4 https://travis-ci.org/ipfs/js-ipfsd-ctl/builds/245233174

each time same test
```
  1 failing
  1) daemons startDaemon starts the daemon and returns valid API and gateway addresses:
     Uncaught AssertionError: expected null to be an instance of Multiaddr
      at async.waterfall (test/index.spec.js:450:52)
      at node_modules/async/dist/async.js:421:16
      at next (node_modules/async/dist/async.js:5278:29)
      at node_modules/async/dist/async.js:906:16
      at node.startDaemon (test/index.spec.js:443:54)
      at f (node_modules/once/once.js:25:25)
      at Socket.data (src/node.js:236:13)
      at readableAddChunk (_stream_readable.js:176:18)
      at Socket.Readable.push (_stream_readable.js:134:10)
      at Pipe.onread (net.js:547:20)
```

diasdavid:
@thisconnect fixed on https://github.com/ipfs/js-ipfsd-ctl/pull/160/files#r135794850

PR #160, opened 2017-08-29
0.4.10

diasdavid:
Having a deja vu seeing the exec tests failing. Last time was an update on `is-running`, now I'm not sure (yet). Errors are:

```
  1) exec SIGTERM kills hang:
     Uncaught AssertionError: expected null to exist
      at exec (test/exec.spec.js:66:22)
      at f (node_modules/once/once.js:25:25)
      at Object.done (src/exec.js:25:9)
      at Stream.listeners.done.once (src/exec.js:45:18)
      at Stream.f (node_modules/once/once.js:25:25)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:200:12)

  2) exec SIGKILL kills survivor:
     Uncaught AssertionError: expected false to be truthy
      at psExpect (test/exec.spec.js:117:31)
      at Timeout.setTimeout [as _onTimeout] (test/exec.spec.js:25:5)
```

However, this time, after the tests are run, there are no zombie daemons

### DIFF
--- a/test/exec.spec.js
+++ b/test/exec.spec.js
@@ -53,8 +53,10 @@ function makeCheck (n, done) {
   }
 }
 
-// TODO Needs issues with https://github.com/ipfs/js-ipfsd-ctl/pull/160#issuecomment-325669206
-// fixed first. The test vector, tail, is what is causing issues
+// TODO The test vector, `tail` is no longer a good test vector as it is not
+// exiting as it once was when the test was designed
+// - [ ] Need test vector or figure out why tail changed
+// Ref: https://github.com/ipfs/js-ipfsd-ctl/pull/160#issuecomment-325669206
 describe.skip('exec', () => {
   it('SIGTERM kills hang', (done) => {
     const tok = token()
--- a/test/exec.spec.js
+++ b/test/exec.spec.js
@@ -53,7 +53,9 @@ function makeCheck (n, done) {
   }
 }
 
-describe('exec', () => {
+// TODO Needs issues with https://github.com/ipfs/js-ipfsd-ctl/pull/160#issuecomment-325669206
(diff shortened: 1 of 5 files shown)


diasdavid:
Ok, a little bit more of investigation and it seems that the issue with the actual vector test, the `tail` command.

Same issue happens in https://github.com/jbenet/node-subcomandante

test/index.spec.js
```diff
@@ -446,8 +446,8 @@ describe('daemons', () => {
 
         expect(daemon).to.have.property('apiAddr')
         expect(daemon).to.have.property('gatewayAddr')
-        expect(daemon.apiAddr).to.be.instanceof(multiaddr)
-        expect(daemon.gatewayAddr).to.be.instanceof(multiaddr)
+        expect(multiaddr.isMultiaddr(daemon.apiAddr)).to.equal(true)
+        expect(multiaddr.isMultiaddr(daemon.gatewayAddr)).to.equal(true)
```
diasdavid: This fixes @thisconnect issue reported on https://github.com/ipfs/js-ipfsd-ctl/issues/158

Note for the future. Using `instanceof` in JS land is bad practice, you can never count on getting the right class/prototype at the right type because npm will install multiple versions (thankfully).

test/exec.spec.js
```diff
@@ -53,7 +53,9 @@ function makeCheck (n, done) {
   }
 }
 
-describe('exec', () => {
+// TODO Needs issues with https://github.com/ipfs/js-ipfsd-ctl/pull/160#issuecomment-325669206
```
VictorBjelkholm: Don't understand this comment (or the linked one either really), could we write down what's needed to make these tests pass?

diasdavid: Updated the comment to present a clear action item.

PR #164, opened 2017-09-04
feat: replaces path to .asar.unpacked if cmd includes .asar/node_modu…

diasdavid:
@thisconnect moved this here so that I can collaborate directly with you on the same branch


diasdavid:
image

PR #165, opened 2017-09-06
Slay the Dragon!

diasdavid:
It seems that **running our tests in Node.js 8 always fail in AppVeyor and locally** (not in Travis or Circle CI interestingly enough)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,7 +3,7 @@
 const os = require('os')
 const join = require('path').join
 
-const Node = require('./node')
+const Node = require('./daemon')
 
 const defaultOptions = {
   'Addresses.Swarm': ['/ip4/0.0.0.0/tcp/0'],
--- a/src/index.js
+++ b/src/index.js
@@ -5,6 +5,7 @@ const join = require('path').join
 
 const Node = require('./daemon')
 
+// Note how defaultOptions are Addresses.Swarm and not Addresses: { Swarm : <> }
 const defaultOptions = {
   'Addresses.Swarm': ['/ip4/0.0.0.0/tcp/0'],
   'Addresses.Gateway': '',
@@ -33,6 +34,7 @@ const IpfsDaemonController = {
   version (callback) {
     (new Node()).version(callback)
   },
+
   /**
    * Create a new local node.
    *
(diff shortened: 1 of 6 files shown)


test/index.spec.js
```diff
@@ -467,34 +467,32 @@ describe('daemons', () => {
...
 
     it('starts the daemon and returns valid API and gateway addresses', (done) => {
-      let daemon
       const dir = `${os.tmpdir()}/tmp-${Date.now() + '-' + Math.random().toString(36)}`
 
       async.waterfall([
         (cb) => ipfsd.local(dir, cb),
-        (node, cb) => {
-          daemon = node
-          node.init((err) => cb(err, node))
```
diasdavid: Seems like in Node.js 8 a copy was being made instead of a reference.. WAT?

dignifiedquire: Hmm that sounds dubious, we should change the test, if this is making copies there is likely a larger issue at play which will be a problem for users of ipfsd-ctl

diasdavid: It seems that this didn't fixed it after all, I just got lucky

diasdavid: got it

diasdavid:
This PR solves the issues that https://github.com/ipfs/js-ipfsd-ctl/pull/150 was trying to solve.

diasdavid:
Given the _sensitivity_ of this module, I appreciate all the CR I can get :)

tl;dr; There were some big issues on how daemons were being spawned and it mostly worked because the race condition was in our favor. Node.js 8 changed that.

diasdavid:
image

wubbalubbadubdub 🎉

test/spawning-daemons.spec.js
```diff
@@ -467,34 +411,32 @@ describe('daemons', () => {
...
-        expect(res).to.have.property('gatewayPort')
-        expect(res.apiHost).to.equal('127.0.0.1')
-        expect(res.apiPort).to.equal('5001')
-        expect(res.gatewayHost).to.equal('127.0.0.1')
-        expect(res.gatewayPort).to.equal('8080')
+        // Check for props in ipfs-api instance
+        expect(api).to.have.property('apiHost')
+        expect(api).to.have.property('apiPort')
+        expect(api).to.have.property('gatewayHost')
+        expect(api).to.have.property('gatewayPort')
```
VictorBjelkholm: These checks are unnecessary as the ones below will check if the property exists and is equal to the value.

src/index.js
```diff
@@ -109,6 +91,29 @@ const IpfsDaemonController = {
     } else {
       node.init((err) => callback(err, node))
     }
+  },
+
+  /**
+   * Create a new disposable node and already start the daemon.
```
VictorBjelkholm: Minor typo
> already started daemon

diasdavid:
Thanks @VictorBjelkholm :)

PR #179, opened 2017-12-20
fix: init on different path

hacdias:
This fixes a small bug where you couldn't init a repository in a different directory than HOME because `this.env` wasn't being updated. I decided to make it a getter so you wouldn't need to change it every time you changed the path.


diasdavid:
@dryajov merge this one if it looks good to you. Fixes one issue @hacdias had

PR #184, opened 2018-01-18
make sure exec matches the passed node type

dryajov:


dryajov:
@diasdavid can we get this one merged, its needed for https://github.com/ipfs/interop/pull/5

package.json
```diff
@@ -93,6 +92,7 @@
     "dirty-chai": "^2.0.1",
     "go-ipfs-dep": "0.4.13",
     "ipfs": "^0.27.5",
+    "ipfs-api": "^17.3.0",
```
diasdavid: @dryajov only realized after that ipfs-api is now a devDep. Does this mean that folks have to pass ipfs-api everytime they want to spawn a node?

diasdavid: It continues to be a prod dep https://github.com/ipfs/js-ipfsd-ctl/blob/master/src/daemon-node.js#L5

diasdavid: What was the rationale behind moving it?

Issue #188, opened 2018-01-27
Make testing fast again - Bring init options back


richardschneider:
Spawning a deamon on my machine (windows) is extremely slow.  It appears that each go/js command that is run takes ~10 seconds!

More details to follow.

dryajov:
10s is definitely a little slow, but keep in mind that each spawned daemon (unless pointing to an existing repo) will have to init the repo and apply any custom configs before launching the damon, which could get pretty slow. But please do provide details, I'd like to speed this up as much as possible 👍

richardschneider:
Its not just `ipfs init`.  Been sided tracked with other issues, but should have details in an hour or so.

FWI: Timing was done by hand (one onesecond, two one-second, ...) and each command (init, config) took exactly 10s.

Also, are the config settings applied to in-proc IPFS?

diasdavid:
@dryajov when the new version of ipfsd-ctl, we lost the ability to set key size on init. This is really unfortunate as tests are taking forever nowadays.

My proposed solution: 
<img width="805" alt="image" src="https://user-images.githubusercontent.com/1211152/36253404-bd767888-1247-11e8-8bc3-1dc0af0349cb.png">

init should be a bool or object and the object should accept any of the init options:

<img width="510" alt="image" src="https://user-images.githubusercontent.com/1211152/36253430-d76c1db0-1247-11e8-8bf2-4351a69812d0.png">

richardschneider:
@diasdavid what about the PR [smaller keysize when testing](https://github.com/ipfs/js-ipfsd-ctl/pull/198)?

diasdavid:
@richardschneider we are not going to patch globals. This should be an option and not a surprise.

dryajov:
> @dryajov when the new version of ipfsd-ctl, we lost the ability to set key size on init. This is really unfortunate as tests are taking forever nowadays.

We didn't, you can still set it as part of calling `ipfsd.init({keysize: <size>})`. The `init` flag was always a `bool` - 
https://github.com/ipfs/js-ipfsd-ctl/blob/v0.26.0/src/index.js#L9..L15.

That said, I agree it's a valuable option to have. In order to not break the existent implementation, lets leave the `init` as a bool, but add a new property, `initOpts` that would take a `keysize`, that would also support any future init option that we might require.

diasdavid:
[...]

PR #190, opened 2018-01-27
Make spawning super fast!

richardschneider:
@dryajov Some details for #188 

Well it was having a bad night, the 10s/command is now down to ~1.5s/command.  It's still slow.  

I'll be very interested in seeing what unix reports.  To get these results, you need to set the environment variable, in windows its `set DEBUG=ipfsd-ctl*`. Then `npm run test:node`.

```
  ipfsd-ctl:run ipfs.exe init -b 2048 +3s
  ipfsd-ctl:run ipfs.exe config API.HTTPHeaders.Access-Control-Allow-Origin ["*"] --json +2s
  ipfsd-ctl:run ipfs.exe config API.HTTPHeaders.Access-Control-Allow-Methods ["PUT","POST","GET"] --json +1s
  ipfsd-ctl:run ipfs.exe config Addresses.Swarm ["/ip4/127.0.0.1/tcp/0"] --json +1s
  ipfsd-ctl:run ipfs.exe config Addresses.API "/ip4/127.0.0.1/tcp/0" --json +1s
  ipfsd-ctl:run ipfs.exe config Addresses.Gateway "/ip4/127.0.0.1/tcp/0" --json +1s
          √ create node and init (7084ms)
  ipfsd-ctl:run ipfs.exe daemon +1s
          √ start node (3790ms)```

### CODE REVIEW

**Comment by richardschneider** [comment]
@diasdavid @dryajov `ipfsd-ctl` uses `subcommandte` to run the command.  It creates another NodeJS process that spawns the actual command.  Seem very inefficient to me.  `subcommandte` is fine when a long running child process needs to be killed when the parent the process is killed.  However, this is not our case, our child process are short lived and wait for them to close.

I'll try calling `child_process` directly and see what happens.

**Comment by richardschneider** [comment]
It's now ~350ms/command.

```
        spawn an initialized node
  ipfsd-ctl:run ipfs.exe init -b 2048 +2s
  ipfsd-ctl:run ipfs.exe config API.HTTPHeaders.Access-Control-Allow-Origin ["*"] --json +1s
  ipfsd-ctl:run ipfs.exe config API.HTTPHeaders.Access-Control-Allow-Methods ["PUT","POST","GET"] --json +303ms
  ipfsd-ctl:run ipfs.exe config Addresses.Swarm ["/ip4/127.0.0.1/tcp/0"] --json +286ms
  ipfsd-ctl:run ipfs.exe config Addresses.API "/ip4/127.0.0.1/tcp/0" --json +289ms
  ipfsd-ctl:run ipfs.exe config Addresses.Gateway "/ip4/127.0.0.1/tcp/0" --json +291ms
          √ create node and init (2584ms)
  ipfsd-ctl:run ipfs.exe daemon +288ms
```

**Comment by diasdavid** [comment]
@richardschneider historically, we had many issues with ipfs process spawning and killing, to the point that using `ipfs-desktop` would spawn dozens of IPFS nodes without being able to effectively shut them down.

In order to move away from subcomandante, we need to be extremely careful and test well that we don't end up in that world again.

**Comment by richardschneider** [comment]
@diasdavid "Test well", does that mean that all tests run on all platforms?

**Comment by diasdavid** [comment]
More than unit tests. I'm talking about linking it to apps like ipfs-desktop and using it in multiple platforms + run the js-ipfs-api tests with it to ensure that everything is running as expected.

**Comment by richardschneider** [comment]
@diasdavid ran the `js-ipfs-api` tests in node and except for the 2 known pubsub test issues, everything passed.  I also checked for zombie processes and none were found.  During the test, at most only 4 js/go processes were active.

Then tried running the tests in the browser and as you can guess `webpack` failed with `heap error`.

I don't know about `ipfs-desktop`. How should I go about testing it?

File: package.json [hunk 1]
```diff
@@ -67,9 +67,11 @@
   "dependencies": {
     "async": "^2.6.0",
     "boom": "^7.1.1",
+    "debug": "^3.1.0",
     "detect-node": "^2.0.3",
     "hapi": "^16.6.2",
     "hat": "0.0.3",
+    "ipfs-api": "^17.3.0",
```
dryajov: can we keep this as a devDependency? We don't want to bundle this, in case the consuming project already defines their own version.

dryajov:
@richardschneider I wonder why such a performance bust? I'd imagine go being a bit lighter than node in general?

package.json
richardschneider: Sorry about that.  It seems that `npm install` just does that.

richardschneider:
@dryajov I beginning to feel that killing a `node subprocess` is causing the issue, especially on Windows.

I'm fighting other fires right now, so will eventually get back to this.

diasdavid:
> I don't know about ipfs-desktop. How should I go about testing it?

@hacdias can you give a hand here?

richardschneider:
@diasdavid I've been waiting for `ipfsd-ctl` to settle down before continuing with this PR.

I'll rebase real soon and continue my investigations.

diasdavid:
@richardschneider thank you :)

diasdavid:
@richardschneider wanna rebase now?

richardschneider:
@diasdavid I need [graceful stop](https://github.com/ipfs/js-ipfsd-ctl/pull/205) before rebasing. What does another week matter.

diasdavid:
@richardschneider time to rebase master onto this branch! :D

diasdavid:
I've rebased master onto this branch, here are my results:

```
# master
npm test  159.45s user 27.56s system 65% cpu 4:45.73 total
# perf
npm test  134.43s user 21.23s system 63% cpu 4:06.89 total
```

dryajov:
This seems to work as expected. Should we go with it (I'm 👍)?

@richardschneider if we merge this, lets not forget to remove subcomandante from the package.json deps

richardschneider:
@dryajov done!

diasdavid:
Please take into account my comment above -- https://github.com/ipfs/js-ipfsd-ctl/pull/190#issuecomment-361034791 -- simply trusting CI won't do it for this PR.

dryajov:
@hacdias how can we test this with ipfs-desktop?

travisperson:
**TL;DR;** - I think we need to look into this further to make sure we are handling things correctly and allowing gracefully shutdown across all platforms, while also being a friendly module to users consuming `ipfsd-ctl`.

---

I believe it is generally frowned upon to listen to the main process signals in a library. The issue is that when listening on `SIGINT` and `SIGTERM` the default node handlers are removed that cause the process to exit. Any module / application must implement its own handlers if it wishes to exit from the signals when the event loop is not empty if it consumes this module.

The `subcomandante` module gets around this as it does not listen on these signals from the consumed module, but from the forked child `subcom`, and polls for the parent `PID`.

It's a little dense and hard to parse, but the section on [Signal Events](https://nodejs.org/api/process.html#process_signal_events) provides some insight to how nodejs signals are handled.

I think it's also important to note this line from the process docs with regard to windows

> Sending SIGINT, SIGTERM, and SIGKILL cause the unconditional termination of the target process.

The standard `ChildProcess.prototype.kill` sends a `SIGTERM`, which cannot be handled on windows, `subcomandante` doesn't really handle this well either. It will turn a `SIGKILL` into a `SIGHUP` though.

Both nodejs and golang exit the process on a `SIGHUP`.

_Side note of Windows_
> `SIGHUP` is generated on Windows when the console window is closed [...] however, Node.js will be unconditionally terminated by Windows about 10 seconds later

The `go-ipfs` project gracefully handles the [`SIGHUP`](https://github.com/ipfs/go-ipfs/blob/master/cmd/ipfs/main.go#L408). However, `js-ipfs` [does not](https://github.com/ipfs/js-ipfs/blob/master/src/cli/commands/daemon.js#L54,L55) and instead will immediately terminate.

I don't have a good solution to this at the moment, but I do think we need to dig a bit further to see if we can come up with a better solution.

---

This thread (https://github.com/nodejs/node/issues/12378) on the node project provides more insight as well, and is worth a read through.

Signals in golang can also be read up on (https://golang.org/pkg/os/signal) here.

dryajov:
Thanks @hacdias!

dryajov:
I agree, we need to look into this further. I think that by removing the intermediate subcomandante process we fall back into the same issues subcomandante is trying to mitigate - namely forcefully killing ipfsd-ctl will leave a zombi ipfs process.

We should definitely handle the `SIGHUP` signal in js-ipfs, and it even works on windows - **amazing** 👅 !

diasdavid:
I'm closing this PR given that the risk / speed tradeoff doesn't justify spending much more time into it right now.

PR #192, opened 2018-01-28
docs: fix disposable default

dryajov:


dryajov:
@richardschneider this is some testing I was doing trying to figure out why CI was failing overall. I don' think we need this - I'll revert.

Issue #193, opened 2018-01-28
Debug logging


richardschneider:
If debug logging is enabled,  via DEBUG environment variable, then all the `js-ipfs` tests fails.  It appears that `ipfsd-ctl` does not like debug output.

```
> set DEBUG=*
> aegir test -t node -f test\http-api\interface\block
```

See the [debug.txt](https://github.com/ipfs/js-ipfsd-ctl/files/1671711/debug.txt)

dryajov:
@richardschneider see - https://github.com/ipfs/js-ipfsd-ctl/pull/195#issuecomment-361124921

In esense, if we enable `DEBUG=*` or `DEBUG=ipfs`, etc - it will enable the daemon debug output, which will break our `stdout/stderr` logging. We can use debugging, but we have to be careful to scope it correctly.

richardschneider:
With so many processes running in a test, I really need debug logging to figure out why the tests are failing. 

ipfsd `stdout/stderr` parsing **SHOULD** work with debug logging.

dryajov:
I'd love to get that working, I just don't see how. Once you enable debugging, it will clog `stderr` and will have to workaround parsing leggit errors from conflicting log messages.

dryajov:
This should be done with https://github.com/ipfs/js-ipfsd-ctl/pull/195

PR #195, opened 2018-01-29
feat: debug logging

richardschneider:
Just some simple debug statements to figure out what is happening.


dryajov:
@richardschneider thanks! We have to be careful to scope debug output correctly and only use that scope when debugging, otherwise it will turn `ipfs` logging, which will break our stdout/stderr parsing.

richardschneider:
@dryajov not sure what you are saying?  

These are debug statements for  `ipfsd-ctl`. it has nothing to do with issue #193

package.json
```diff
@@ -67,9 +67,11 @@
   "dependencies": {
     "async": "^2.6.0",
     "boom": "^7.1.1",
+    "debug": "^3.1.0",
     "detect-node": "^2.0.3",
     "hapi": "^16.6.2",
     "hat": "0.0.3",
+    "ipfs-api": "^17.3.0",
```
dryajov: this needs to be moved to devDependencies

diasdavid: @dryajov why devDep and not dep?

richardschneider: @dryajov  Doesn't `ipfsd-ctl` need `ipfs-api` to control the remote daemon?

dryajov: @richardschneider it does, but we've had issues bundling deps like that before (it was `ipfs` proper in that case, but it applies to `ipfs-api` as well).

richardschneider: Is it a circular reference, both need each other?

diasdavid: Why does it apply to `ipfs-api` as well? `ipfs` caused issues because it brought in some native dependencies that were causing trouble for ipfs-desktop. It was ok to move `ipfs` to devDep because we moved it to be injected by the user (the other option being go-ipfs-dep). However, `ipfs-api` is always used and doesn't make sense that a user needs to inject as well.

dryajov: IMO, we should not bundle neither `ipfs` not `ipfs-api` and let the consumer bundle their own, because we can't guarantee which version is going to be ultimately used by `ipfsd-ctl` in either case.

> However, ipfs-api is always used and doesn't make sense that a user needs to inject as well.

So is `js-ipfs`, the consuming project might have a different version bundled already. 

I'm fine with bundling `ipfs-api`, I don't think it's going to cause any issues, but allowing the consuming project use their own, IMO gives it more control. It's not a big issue either way, because the consumer can alway explicitly use their own `ipfs-api` version with `IpfsApi(ipfsd.apiAddr)`.

richardschneider: Hey guys, I just want some debug logging.  Please make a decision.

dryajov: I'm fine including it 👍

diasdavid: > I'm fine with bundling ipfs-api, I don't think it's going to cause any issues, but allowing the consuming project use their own, IMO gives it more control.

What kind of control do you expect? `ipfs-api` is the client library to talk with daemons. 

If I understand correctly, the only reason why we don't pack in `ipfs` and `npm-go-ipfs-dep` by default is due to issues with electron, not because of control.

richardschneider:
@diasdavid are you happy to merge?

diasdavid:
@richardschneider I'm now worried that the lack including `ipfs-api` here was enabling the tests on `js-ipfs-api` to pass because of how require works. Could you double check and test that js-ipfs, js-ipfs-api and `ipfs/interop` tests pass with this version?

richardschneider:
@diasdavid CI tests are failing all over the place.

Running `js-ipfs`, `js-ipfs-api` and `ipfs-interop` will take most of my day and then most likely yield false positive results.  If the code base can not handle adding a few debug statements into a module, then we have **very** serious issues.

I am adding this PR, so I can figure out why tests are running so slow.  Blaming Travis or sprinkling `this.timeout` is just not working.

dryajov:
@richardschneider could you point to what's failing? 

@diasdavid in `js-ipfs-api` we use the current `js-ipfs-api` from `src/` and instantiate it explicitly using `IpfsApi(ipfsd.apiAddr)`. `js-ipfs` and `interop` would pick up the latest `ipfs-api` from their dependencies. Not sure how this would have affected it, specially with later versions of `npm` doing explicit dedup on those dependencies.

richardschneider:
@dryajov See 
- https://circleci.com/gh/ipfs/js-ipfs-api/1655?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
- https://travis-ci.org/ipfs/js-ipfs/jobs/335828903

Different repos and CIs with different types of random errors.

dryajov:
@richardschneider I don't think they are related to this changes, I've seen them come up randomly at different times. 

https://circleci.com/gh/ipfs/js-ipfs-api/1655?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link - I believe its been mentioned somewhere else might be due to webpack.

https://travis-ci.org/ipfs/js-ipfs/jobs/335828903 - has been a consistent random error on travis that pops up now and then, restarting the build fixes it. I'm sure there is a race condition somewhere that might be causing that and we should definitely investigate, but it's been around for a while and predates using `ipfsd-ctl`.

dryajov:
To clarify, and make sure we're on the same page.

- `ipfsd-ctl` was shipped with `ipfs-api` as a `devDependency` in hopes it would cause less conflicts and allow consuming projects use the latests available `ipfs-api` to them without having to update `ipsd-ctl` constantly
- `js-ipfs` is also a `devDependency`, however the initial reason for moving it there was due to the native dependencies that were affecting electron builds, however, we have the same side effect as with `ipfs-api` - i.e. we can use the latest `js-ipfs` in consuming projects without having to ship a new versions of `ipfsd-ctl` every time

The concerns are that, shipping `ipfs-api` as a devDependency could affect testing across the different projects. 

I don't believe that it's the case, because:

- in `js-ipfs-api`, `ipfsd-ctl` uses the `ipfs-api` from its `src/` dir, by explicitly instantiating it with `IpfsApi(ipfsd.apiAddr)`
- in the all other projects, it should default to using the version of `js-ipfs` and `ipfs-api` declared in their `dependencies`

I must be missing something and I would love to see where the issues might lie.

diasdavid:
@dryajov you are missing all the projects that use ipfsd-ctl to spawn go-ipfs daemons that are not js-ipfs nor js-ipfs-api.

dryajov:
> @dryajov you are missing all the projects that use ipfsd-ctl to spawn go-ipfs daemons that are not js-ipfs nor js-ipfs-api.

Fair point, thanks for bringing that up. As I said, I'm fine moving `ipfs-api` back to `dependencies`.

I'm not clear on why it would affect existing tests in `js-ipfs` and `interop` though?

diasdavid:
> I'm not clear on why it would affect existing tests in js-ipfs and interop though?

I'm not saying that it will. I just want us to be diligent and make sure that we test the dependents when we make breaking changes. I've seen multiple times PRs that "ready" to later see _quick fixes_ all over when things start failing on the dependents.

dryajov:
@diasdavid got you! Thanks for clarifying. Agree on being diligent - point taken.

richardschneider:
@dryajov @diasdavid any progress?

richardschneider:
@diasdavid could you just merge this (not release to NPM).  I really need this to test some speedups.

Pretty please with a cherry on the top.

PR #196, opened 2018-01-29
fix: make sure repo is being properly cleaned up

dryajov:


test/api.js
```diff
@@ -16,8 +16,8 @@ const isWindows = os.platform() === 'win32'
 
 module.exports = (df, type) => {
   return () => {
-    const API_PORT = type === 'js' ? '5002' : '5001'
-    const GW_PORT = type === 'js' ? '9090' : '8080'
+    const API_PORT = type === 'js' ? '5102' : '5101'
```
VictorBjelkholm: Not sure this is much better. Can't we use `0` here and later get to port for later use?

dryajov: hmm, I thought about that, but this test makes sure that we're starting the API and Gateway on the specified ports, so in this case I think it makes sense - not sure about checking for the node type tho, I think thats not needed anymore.

richardschneider:
@dryajov this is failing on Travis CI because of a lint error, please fix let's get this merged.
```
/home/travis/build/ipfs/js-ipfsd-ctl/src/daemon-node.js
  244:1  error  More than 1 blank line not allowed  no-multiple-empty-lines
```

Issue #194, opened 2018-01-29
Passphrase when initing


richardschneider:
Need to allow an optional passphrase when initing.

Add `pass: null` to the `initOpts`.  If not null, then produce `jsipfs init --pass ... -b 2048`.

Make sure that this is also done for in-proc deamons.

dryajov:
addressed in - https://github.com/ipfs/js-ipfsd-ctl/pull/187

PR #187, opened 2018-01-27
Various enhancements

richardschneider:
- Passphrase for in-proc node #186
- Running JS on Windows #185

PR #197, opened 2018-01-29
add passphrase to init options

richardschneider:
See #194

### DIFF
--- a/src/daemon-node.js
+++ b/src/daemon-node.js
@@ -108,6 +108,7 @@ class Node {
    * @param {Object} [initOpts={}]
    * @param {number} [initOpts.keysize=2048] - The bit size of the identiy key.
    * @param {string} [initOpts.directory=IPFS_PATH] - The location of the repo.
+   * @param {string} [initOpts.pass] - The passphrase of the keychain.
    * @param {function (Error, Node)} callback
    * @returns {undefined}
    */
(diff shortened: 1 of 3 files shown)


src/daemon-node.js
```diff
@@ -117,13 +118,16 @@ class Node {
...
 
-    const keySize = initOpts.keysize || 2048
-
     if (initOpts.directory && initOpts.directory !== this.path) {
       this.path = initOpts.directory
     }
 
-    run(this, ['init', '-b', keySize], { env: this.env }, (err, result) => {
+    const args = ['init', '-b', initOpts.keysize || 2048]
+    if (initOpts.pass) {
```
dryajov: @richardschneider this should already happen since args just get forwarded to the daemon.

dryajov:
@richardschneider I see we already addressed passing the `--pass` options to in-proc nodes here - https://github.com/ipfs/js-ipfsd-ctl/pull/187.

richardschneider:
@dryajov Yes `--pass` works when spawing the daemon.  This PR is to allow it when initing the repo.

dryajov:
@richardschneider `init()` taks an `args` array, that should allow you to pass the `--pass` option already, so no need for that.

richardschneider:
Seems like my **bad**.  So something like the below will add `-pass` to all commands generated to init.

```
const options = {
  args: ['--pass ipfs-is-awesome-software']
}
df.spawn(options, (err, _ipfsd) => { ... })
```

dryajov:
@richardschneider - actually my bad. You're right, init takes an `initOpts` object, and if `pass` is a flag for `init` then we do need this change.

richardschneider:
@dryajov Don't you have PR to fix the jenkins port sharing issue?  If so can we merge it.

dryajov:
@richardschneider yes, it's here  - https://github.com/ipfs/js-ipfsd-ctl/pull/196. 

Just need to make sure that we don't start non `disposable` nodes with the default config (https://github.com/ipfs/js-ipfsd-ctl/pull/196/files#diff-b7c01f8fb93cfc68f6e35c7af488a039R192).  

i.e. 

use something like this when starting non `disposable` nodes:

```
const config = {
            Addresses: {
              Swarm: [
                '/ip4/127.0.0.1/tcp/0/ws',
                '/ip4/127.0.0.1/tcp/0'
              ],
              API: '/ip4/127.0.0.1/tcp/0',
              Gateway: '/ip4/127.0.0.1/tcp/0'
            }
          }
```

PR #198, opened 2018-02-03
Smaller key size when testing

richardschneider:
A 2048 bit RSA key can take up to 20 seconds to generate.  When testing, a smaller key size can be used.

For `js-ipfs` a 512 bit key size is used.  For `go-ipfs`, because it enforces minimum key size, 1024 bits is used.

### DIFF
--- a/src/daemon-node.js
+++ b/src/daemon-node.js
@@ -106,11 +106,14 @@ class Node {
    * Initialize a repo.
    *
    * @param {Object} [initOpts={}]
-   * @param {number} [initOpts.keysize=2048] - The bit size of the identiy key.
+   * @param {number} [initOpts.keysize=2048] - The bit size of the identity key.
    * @param {string} [initOpts.directory=IPFS_PATH] - The location of the repo.
    * @param {string} [initOpts.pass] - The passphrase of the keychain.
    * @param {function (Error, Node)} callback
    * @returns {undefined}
+   *
+   * If running in a test, then a smaller default keysize is used to
+   * improve speed over security.
    */
   init (initOpts, callback) {
     if (!callback) {
@@ -122,7 +125,16 @@ class Node {
       this.path = initOpts.directory
     }
 
-    const args = ['init', '-b', initOpts.keysize || 2048]
+    let keysize = initOpts.keysize
+    if (!keysize) {
+      if (typeof global.it === 'function') {
+        keysize = this.opts.type === 'go' ? 1024 : 512
+      } else {
+        keysize = 2048
+      }
+    }
+
+    const args = ['init', '-b', keysize]
     if (initOpts.pass) {
       args.push('--pass')
       args.push('"' + initOpts.pass + '"')
--- a/src/in-proc-node.js
+++ b/src/in-proc-node.js
@@ -129,7 +129,7 @@ class Node {
       initOpts = {}
     }
 
-    initOpts.bits = initOpts.keysize || (typeof global.it === 'function') ? 512 : 1024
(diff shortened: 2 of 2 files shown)


src/daemon-node.js
```diff
@@ -122,7 +125,16 @@ class Node {
       this.path = initOpts.directory
     }
 
-    const args = ['init', '-b', initOpts.keysize || 2048]
+    let keysize = initOpts.keysize
+    if (!keysize) {
+      if (typeof global.it === 'function') {
```
dryajov: @richardschneider @diasdavid I'm really not sure we should do this. 

- Why do we have to change this for test only? 
- Can we be sure that this is not going to hide some subtle issues by running with different defaults in tests vs prod?

richardschneider: > Why do we have to change this for test only?

Because small key sizes are unsafe.  If someone is using this interface to create a 'real' peer then proper key sizes should be used.  This is just hack to get tests to run faster.

`libp2p-keychain` and `libp2p-crypto` tests use real key sizes.

> Can we be sure that this is not going to hide some subtle issues by running with different defaults in tests vs prod?

I did think about this and came to the conclusion that because none of the `ipfs` an `ipfs-api` tests specify a key size it should be safe.  But running the tests against this PR is the only sure way.

dryajov: also, checking for `global.it` ties us to mocha and feels a bit iffy...

richardschneider: @dryajov Open to any suggestions.  But the current tests are in mocha.  Worse case, someone writes a new a test in another framework and all they get is bigger key sizes.

dryajov: @richardschneider can we use an ENV variable to set the keysize, that way, in test we can just set it, rather than checking for `global.it`?

i.e.

`IPFS_KEYSIZE=1024 npm test`

richardschneider: Then all the tests would somehow have to set the new ENV variable.  I was looking for a low cost way of speeding up our tests,

dryajov: that's true, maybe this is a change that needs to happen in aegir, where we set an ENV variable signaling that we're in test - `global.it`, IMO would just end up being too brittle. 

i.e. 

`process.env{AEGIR_TEST}`

dryajov: Thinking about it some more, I really think we should just have `IPFS_KEYSIZE_[JS|GO|PROC]`. This carries the most changes, but it's also the most consistent and gives us more flexibility down the road.

dryajov: The reason I say this, is because `ipfsd-ctl` is not used only in tests, and there might be other use cases and applications that can benefit from this functionality as well.

richardschneider: An other app, could simply set the keysize.  I expect an other app, would init an repo only once.

Our tests are an exception, as there are many of creating many repos..

diasdavid: I agree. Let's not peg on globals that are used by specific test frameworks.

If another batch of tests wants to drop down the keysize, then it should do that explicitly rather than having this module making the decision.

src/daemon-node.js
```diff
@@ -122,7 +125,16 @@ class Node {
...
+        keysize = this.opts.type === 'go' ? 1024 : 512
```
dryajov: @richardschneider why are we setting different sizes for go and js, could you please remind me?

richardschneider: RTFM the top comment on this PR

diasdavid:
closing in favor of #203

PR #200, opened 2018-02-07
The big refactor!

diasdavid:
The current ipfsd-ctl is a big mess. There are ~~no~~ only a few browser tests; tests have the setup assert and teardown phases all mixed with sometimes using a describe + before as the assertion; there are multiple files that get loaded in the browser without necessity, this is causing all sorts of problems on other modules https://github.com/ipfs/js-ipfs-api/issues/683; The internals use different names for the things that are exposed (i.e calling daemon-ctrl to the Daemon Factory).

There is still a considerable amount of work to do here. I might not be able to finish this today but I intend to have it tomorrow.


dryajov:
@diasdavid  I'm not sure what you mean by:

> There are no browser tests; 

All tests run in the browser, what other browser tests are missing?

diasdavid:
I take that back, there is indeed a suite that gets run in the browser https://github.com/ipfs/js-ipfsd-ctl/blob/master/test/browser.js#L4

dryajov:
@diasdavid right, the rest of the tests are concerned with daemon specific tests (i.e. repo exist in the filesystem, daemon cleans up correctly when killed, executable is able to resolve correctly, etc... which can't be tested from the browser). 

Thanks for the refactor 👍 I think it cleans up a lot of things.

.aegir.js
```diff
@@ -2,7 +2,8 @@
 
 const createServer = require('./src').createServer
 
-const server = createServer()
+const server = createServer({ port: 25000 })
```
VictorBjelkholm: Can we use port 0 to avoid port collisions?

diasdavid: This is spawned in a separate process by aegir, we would have to create _yet another side channel_ just to pass this port over. Not worth it. For other cases, yes I like port 0 too.

README.md
```diff
@@ -10,33 +10,14 @@
...
-    | |                       +-------                       --------  R  -----|---- Remote Daemon Ctrl   |  |
-    | +-----------------------+   |  +-----|-----------|-----+      |  V  |    |   |                      |  |
-    |                             |        |           |            |  E  |    |   +----------------------+  |
-    | +-----------------------+   |        |           |            |  R  |    |                             |
-    | |       IPFS API        |   |        |           |            +-----+    |   +----------------------+  |
-    | |                       -------------+           |                       |   |     IPFS API         |  |
-    | +-----------------------+   |                    +-----------------------|----                      |  |
-    |                             |                                            |   +----------------------+  |
-    +-----------------------------+                                            +-----------------------------+
-```
```
diasdavid: Needs a new diagram.

README.md
```diff
@@ -265,6 +263,41 @@ electron-packager ./ --asar.unpackDir=node_modules/go-ipfs-dep
...
+    ├── parse-config.js
+    ├── repo
+    │   ├── create-browser.js
+    │   └── create-nodejs.js
+    ├── run.js
+    ├── set-config-value.js
+    └── tmp-dir.js
+
+4 directories, 21 files
+```
```
diasdavid: I'm finally happy with the code structure :)

src/endpoint/routes.js
```diff
@@ -25,8 +25,11 @@ module.exports = (server) => {
     path: '/spawn',
     handler: (request, reply) => {
       const payload = request.payload || {}
-      const ctrl = new CtrlFactory({ type: payload.type })
-      ctrl.spawn(payload.opts, (err, ipfsd) => {
+
+      // TODO: use the ../src/index.js so that the right Factory is picked
```
diasdavid: - [ ] 🌟

src/factory-daemon.js
```diff
@@ -0,0 +1,119 @@
...
+   * @returns {undefined}
+   */
+  version (options, callback) {
+    if (typeof options === 'function') {
+      callback = options
+      options = {}
+    }
+    options = Object.assign({}, options, { type: this.type })
+    // TODO: (1) this should check to see if it is looking for Go or JS
+    // TODO: (2) This spawns a whole daemon just to get his version? There is
```
diasdavid: - [ ] 🌟

src/factory-in-proc.js
```diff
@@ -50,29 +26,32 @@ class DaemonController {
...
-    opts = opts || {}
-    opts.type = this.type
-    const node = new Node(opts)
-    node.version(callback)
+  version (options, callback) {
+    if (typeof options === 'function') {
+      callback = options
+      options = {}
+    }
+    // TODO create this method
```
diasdavid: - [ ] 🌟

src/ipfsd-daemon.js
```diff
@@ -323,7 +327,8 @@ class Node {
       ),
       (config, cb) => {
         if (!key) {
-          return tryJsonParse(config, cb)
+          // TODO: this function didn't exist on the original codebase
+          // return tryJsonParse(config, cb)
```
diasdavid: - [ ] Figure out why this was here in the first place.

test/start-stop.node.js
```diff
@@ -0,0 +1,280 @@
...
+          repoPath = ipfsd.path
+          done()
+        })
+      })
+
+      it('should return a node', () => {
+        expect(ipfsd).to.exist()
+      })
+
+      // TODO fix this test
```
diasdavid: - [ ] 🌟

test/start-stop.node.js
```diff
@@ -0,0 +1,280 @@
...
+      })
+
+      it('repo should cleaned up', () => {
+        expect(fs.existsSync(repoPath)).to.not.be.ok()
+      })
+
+      it('.start with flags', (done) => {
+        // TODO js-ipfs doesn't fail on unrecognized args. Think what should be
+        // the desired behaviour
+        if (type === 'js') { return this.skip() }
```
diasdavid: - [ ] 🌟

README.md
diasdavid: Here for reference:

```
src
├── defaults
│   ├── config.json
│   └── options.json
├── endpoint                    # endpoint to support remote spawning
│   ├── routes.js
│   └── server.js
├── factory-client.js           # IPFS Factories: client (remote), daemon (go or js) and in-proc (js)
├── factory-daemon.js
├── factory-in-proc.js
├── index.js
├── ipfsd-client.js             # ipfsd (Daemon Controller): client (remote), daemon (go or js), in-proc (js)
├── ipfsd-daemon.js
├── ipfsd-in-proc.js
└── utils                       # Utils used by the Factories and Daemon Controllers
    ├── configure-node.js
    ├── exec.js
    ├── find-ipfs-executable.js
    ├── flatten.js
    ├── parse-config.js
    ├── repo
    │   ├── create-browser.js
    │   └── create-nodejs.js
    ├── run.js
    ├── set-config-value.js
    └── tmp-dir.js

4 directories, 21 files
```

src/factory-daemon.js
dryajov: just doing `new Daemon` won't actually start the daemon, it will simply allow you to send commands to the selected executable.

src/factory-in-proc.js
```diff
@@ -3,44 +3,20 @@
 const defaults = require('lodash.defaultsdeep')
```
dryajov: why duplicate if factory-in-proc share %90+ code with factory-daemon?

src/ipfsd-daemon.js
dryajov: because the daemon process will return a json string back, not an object.

test/add-retrieve.spec.js
```diff
@@ -0,0 +1,63 @@
+/* eslint-env mocha */
```
dryajov: This was originally intended as a set of reusable tests to run after the different combinations of `spawn/init/start` to make sure the daemon were being created in a functional way - this intent is lost with this change. How are we testing for this cases?

README.md
dryajov: What is the current diagram missing?

test/spawn-options.node.js
```diff
@@ -0,0 +1,360 @@
...
+    const VERSION_STRING = dfOpts.type === 'js'
+      ? `js-ipfs version: ${require('ipfs/package.json').version}`
+      : 'ipfs version 0.4.13'
+    let df
+
+    before(() => {
+      df = DaemonFactory.create(dfOpts)
+    })
+
+    // TODO: why does ipfsd-ctl polute the env variables?
```
dryajov: It used to set IPFS_PATH and other variables, its not the case anymore - we can remove that.

test/spawn-options.node.js
```diff
@@ -0,0 +1,360 @@
...
+    })
+
+    describe('init and start manually', function () {
+      let ipfsd
+
+      it('df.spawn', (done) => {
+        const options = {
+          init: false,
+          start: false,
+          disposable: true
```
dryajov: this needs to be false

test/spawn-options.node.js
```diff
@@ -0,0 +1,360 @@
...
+          }),
+          (cb) => {
+            ipfsd.init(cb)
+          },
+          (cb) => {
+            ipfsd.start(cb)
+          }
+        ], (err) => {
+          expect(err).to.not.exist()
+          expect(fs.existsSync(repoPath)).to.be.ok()
```
dryajov: >   // TODO why wouldn't this work from the browser if we use the
     // remote endpoint?


because we use `fs.existsSync` - is there a way of making it work in the browser?

src/ipfsd-daemon.js
diasdavid: @dryajov even if that is correct, the `tryJsonParse` function doesn't seem to exist anywhere in the codebase

dryajov:
@diasdavid I took on fixing some of the outstanding todos as well as making sure all tests pass in node.

src/factory-in-proc.js
diasdavid: We should have a base class then. The main reason is that you want to avoid having factory-daemon being loaded in the browser (and all of its dependencies).

diasdavid:
The last failing test is because the tests are assuming that there is an IPFS repo initialized on the home dir, when things get spawn from factory-client. This means that the custom path is not getting propagated all the way through and it is using the default one.

```
  1) Spawn options,go,init and start,spawn from a initialized repo
       start node:
     Error: Uncaught AssertionError: expected [Error: ENOENT: no such file or directory, open '/home/ubuntu/.ipfs/config'] to not exist (test/spawn-options.spec.js:65496)

  2) Spawn options,js,init and start,spawn from a initialized repo
       start node:
     Error: Uncaught AssertionError: expected [Error: ENOENT: no such file or directory, open '/home/ubuntu/.ipfs/config'] to not exist (test/spawn-options.spec.js:65496)
```

README.md
```diff
@@ -10,33 +10,14 @@
...
-    +-----------------------------+                                            +-----------------------------+
-```
+> Spawn IPFS daemons using JavaScript in Node.js or in the Browser.
 
 ## Table of Contents
 
 - [Install](#install)
 - [Usage](#usage)
 - [API](#api)
+- [Packageing](#packaging)
```
diasdavid: - [ ] typo

src/factory-client.js
```diff
@@ -0,0 +1,77 @@
...
+   * @param {Object} [options={}]
+   * @param {function(Error, string)} callback
+   * @returns {undefined}
+   */
+  version (options, callback) {
+    if (typeof options === 'function') {
+      callback = options
+      options = {}
+    }
+    // TODO implement this method
```
diasdavid: - [ ] This method needs to be implemented as well for clients. Once it is, enable tests again.

src/factory-daemon.js
diasdavid: It still should check if it is JS or Go

src/factory-client.js
dryajov: In progress, will push shortly.

src/factory-daemon.js
dryajov: Why? All the logic around daemon type is already being handled in the `Daemon` itself - not sure what to do with the check there.

diasdavid:
Ok, fixed some of the tests, now I'm seeing a new issue. It seems that when spawning a js-ipfs daemon and/or in-proc node, it doesn't respect the default config of ipfsd-ctl and uses the default of js-ipfs causing port clashes

image

src/factory-in-proc.js
```diff
@@ -101,9 +101,12 @@ class FactoryInProc {
...
       : true
 
-    if (!options.disposable) {
+    if (options.disposable) {
+      options.config = defaults({}, options.config, defaultConfig)
+    } else {
       const nonDisposableConfig = clone(defaultConfig)
-      delete nonDisposableConfig.Addresses
+      // TODO why delete the addrs here???
+      // delete nonDisposableConfig.Addresses
```
diasdavid: @dryajov why were you deleting the addrs here?

src/factory-daemon.js
```diff
@@ -26,7 +26,7 @@ class FactoryDaemon {
     if (options && options.type === 'proc') {
       throw new Error('This Factory does not know how to spawn in proc nodes')
     }
-    options = Object.assign({ type: 'go' }, options)
+    options = Object.assign({}, { type: 'go' }, options)
```
diasdavid: @dryajov why this change?

src/factory-in-proc.js
```diff
@@ -34,6 +37,21 @@ class FactoryInProc {
...
+   * *Here for completeness*
+   *
+   * @param {string} type - the type of the node
+   * @param {function(Error, string)} callback
+   * @returns {undefined}
+   */
+  tmpDir (type, callback) {
+    callback(null, tmpDir(type === 'js'))
+  }
+
```
diasdavid: @dryajov you are assuming that a Node in Proc only exists in Node.js.. but it can also be browser.

src/factory-in-proc.js
```diff
@@ -4,11 +4,14 @@ const defaults = require('lodash.defaultsdeep')
 const clone = require('lodash.clone')
 const waterfall = require('async/waterfall')
 const path = require('path')
+const tmpDir = require('./utils/tmp-dir')
```
diasdavid: This brought the os module to the browser.

diasdavid:
image

Woot! Good to merge. Deferring remaining tune up to later PRs

src/factory-in-proc.js
dryajov: Because at some point we wanted non-disposable nodes to use `js-ipfs` or `go-ipfs` default ports and not random ports.

PR #203, opened 2018-02-16
feat: add keysize through an option to spawn

dryajov:
wip, tests are on the way


dryajov:
ref #188

README.md
```diff
@@ -115,7 +115,8 @@ Install one or both of the following modules:
 Spawn the daemon
 
 - `options` is an optional object the following properties:
-  - `init` bool (default true) - should the node be initialized
+  - `init` bool (default true) or Object - should the node be initialized
+    - if `init` is an Object, it is expected to be of the form `{keySize: <bits>}`, this will set the `keysize` to initialize the daemon with
```
VictorBjelkholm: Not a fan of conditional types in a parameter. Either it's a `Boolean` or it's a `Object`. Rather, we should add a new option. If we're only adding `keySize`, why not make `keySize` a key of the `options` object?

src/ipfsd-daemon.js
```diff
@@ -57,6 +57,14 @@ class Daemon {
     this._gatewayAddr = null
     this._started = false
     this.api = null
+    this.keySize = null
+
+    this.keySize = process.env.IPFS_KEYSIZE
+
+    // option takes precedence over env variable
```
VictorBjelkholm: I don't think this is right. First it should use the default value. Second it should use the passed in value and lastly, if there is a environment variable, use that. 

This way we can override application configs when deploying etc.

README.md
VictorBjelkholm: I also think we call the option `bits` here, OR, change the option in go-ipfs as well. It's `bits` there currently, but `keySize` is probably a more descriptive name.

dryajov: Thanks @VictorBjelkholm, I thought about that too and you're right, I'll add another options `initOpts` and leave `init` as a bool.

For the second comment, `keysize` is used because historically that is what `ipfsd-ctl` used and it's the same name we use when passing that flag to `init`, and yes I think `keysize` is more descriptive.

diasdavid: read: https://github.com/ipfs/js-ipfsd-ctl/issues/188#issuecomment-366299208

I'm fine with init being boolean or object, but I'm fine with initOpts too.

src/ipfsd-daemon.js
diasdavid: 👍 for what @VictorBjelkholm said.

diasdavid:
@dryajov while at this, update the tests to use smaller keysizes as well so that they run faster :)

src/ipfsd-daemon.js
dryajov: Wait, what @VictorBjelkholm said reads as - default take precedence over env, which takes precedence over option? 

Seems backwards, unless I'm reading it wrong. In any case, this is what I'm doing now - option takes precedence over ENV variable, which takes precedence over default. I think that is the canonical way of doing it and allows passing defaults with the ENV variable, but override them on a case by case basis in tests cases and elsewhere?

diasdavid: `Default -> Options trumps Default -> Env trumps Options and Default`

diasdavid: @dryajov I believe you are using `precedence` as "it comes before === less value" but the english definiton says "precedence: the condition of being considered more important than someone or something else; priority in importance, order, or rank." and so, if default took precedence over env, default values would always win.

dryajov: > option takes precedence over ENV variable, which takes precedence over default.

Which uses `precedence` in the english definition of the term.

I think the contention is whether ENV take precedence over option or vice versa. The case I'm making is that an ENV variable will define a global that should be overridable through an option when calling `spawn`. This allows to override the default globally and locally on a case by case basis.

dryajov: So in my case it is -

`Default -> Env trumps Default -> Options trumps Default and Env`

But I'm not dead set on it, but it feels more natural to me.

diasdavid: @dryajov use https://github.com/ipfs/js-ipfsd-ctl/pull/203#discussion_r168822236, which is pretty much how every system works. 

I wonder which system are you familiar with that uses it the other way around.

dryajov: Any programming language with correct scoping rules :)

dryajov:
@diasdavid @VictorBjelkholm btw, any quick way of verifying that the keys were generated with the correct size?

diasdavid:
@dryajov not a deterministic one, other than a smaller key in bit size should be smaller in storage size. see https://github.com/libp2p/js-peer-id/blob/master/test/peer-id.spec.js#L124-L134

test/start-stop.node.js
```diff
@@ -212,6 +211,9 @@ types.forEach((type) => {
     })
 
     describe('start and stop multiple times', () => {
+      // TODO: wont work on windows until we get /shutdown implemented in js-ipfs
```
richardschneider: It's now working, see PR https://github.com/ipfs/js-ipfsd-ctl/pull/205

dryajov: Nice... is it ready to review and merge?

richardschneider: waiting for a review by @diasdavid 

My PR has a link to js-ipfs/master (not using npm) so it can be used immediately.

package.json
```diff
@@ -7,13 +7,13 @@
...
-    "test:node": "aegir test -t node",
-    "test:browser": "aegir test -t browser --no-cors",
+    "test": "IPFS_KEYSIZE=1024 aegir test -t node -t browser --no-cors",
+    "test:node": "IPFS_KEYSIZE=1024 aegir test -t node",
+    "test:browser": "IPFS_KEYSIZE=1024 aegir test -t browser --no-cors",
     "release": "aegir release",
     "release-minor": "aegir release --type minor",
     "release-major": "aegir release --type major",
-    "coverage": "COVERAGE=true aegir coverage --timeout 50000",
+    "coverage": "IPFS_KEYSIZE=1024 COVERAGE=true aegir coverage --timeout 50000",
```
diasdavid: @dryajov please set these values on tests themselves.

src/factory-daemon.js
```diff
@@ -74,8 +74,8 @@ class FactoryDaemon {
    * Spawn an IPFS node, either js-ipfs or go-ipfs
    *
    * Options are:
-   * - `init` {bool|Object} - should the node be initialized
-   *  - if `init` is an Object, it is expected to be of the form `{keySize: <bits>}`
+   * - `init` bool - should the node be initialized
+   * - `initOpts` Object, it is expected to be of the form `{bits: <size>}`, which sets the desired key size
```
diasdavid: Update the README as well

src/ipfsd-daemon.js
```diff
@@ -57,14 +57,9 @@ class Daemon {
...
-    this.keySize = null
+    this.bits = null
 
-    this.keySize = process.env.IPFS_KEYSIZE
-
-    // option takes precedence over env variable
-    if (typeof this.opts.init === 'object') {
-      this.keySize = this.opts.init.keySize
-    }
+    this.bits = this.opts.initOpts ? this.opts.initOpts.bits : process.env.IPFS_KEYSIZE
```
diasdavid: L60 is redundant

README.md
diasdavid: README needs to be updated to include `initOptions`

src/ipfsd-daemon.js
```diff
@@ -57,6 +58,7 @@ class Daemon {
     this._gatewayAddr = null
     this._started = false
     this.api = null
+    this.bits = this.opts.initOpts ? this.opts.initOpts.bits : process.env.IPFS_KEYSIZE
```
diasdavid: `s/initOpts/initOptions/g`

src/ipfsd-daemon.js
```diff
@@ -128,24 +130,32 @@ class Daemon {
       this.path = initOpts.directory
     }
 
-    const args = ['init', '-b', initOpts.keysize || 2048]
+    const bits = initOpts.bits ? initOpts.bits : this.bits
```
diasdavid: Isn't this repeating part of line 61?

src/ipfsd-daemon.js
```diff
@@ -128,24 +130,32 @@ class Daemon {
...
+    const args = ['init']
+    // do not just set a default keysize,
+    // in case we decide to change it at
+    // the daemon level in the future
```
diasdavid: This sentence doesn't read right. Do you mean "Skip if default daemon keysize"?

src/ipfsd-daemon.js
```diff
@@ -128,24 +130,32 @@ class Daemon {
...
-      configureNode(this, this.opts.config, (err) => {
-        if (err) {
-          return callback(err)
-        }
-
+      waterfall([
+        (cb) => this.getConfig(cb),
+        (conf, cb) => this.replaceConfig(defaults({}, this.opts.config, conf), cb)
+      ], (err) => {
+        if (err) { return callback }
```
diasdavid: the callback needs to be called

src/ipfsd-daemon.js
```diff
@@ -128,24 +130,32 @@ class Daemon {
...
         this.clean = false
         this.initialized = true
-        callback(null, this)
+        return callback(null, this)
```
diasdavid: Init should not need to return a reference to the instance.

src/ipfsd-daemon.js
```diff
@@ -348,6 +358,33 @@ class Daemon {
...
+   * @param {object} config
+   * @param {function(Error)} callback
+   * @return {undefined}
+   */
+  replaceConfig (config, callback) {
+    const tmpFile = path.join(os.tmpdir(), hat())
+    // I wanted to use streams here, but js-ipfs doesn't
+    // read from stdin when providing '-' (or piping) like
+    // go-ipfs, and adding it right now seems like a fair
+    // bit of work, so we're using tmp file for now - not ideal...
```
diasdavid: PR with a WIP here https://github.com/ipfs/js-ipfs/pull/785 really close to finish.

src/ipfsd-daemon.js
dryajov: Currently, ipfs daemons have a default keysize when using `ipfs init` - if we pass a value with `ipfs init -b <keysize>`, we're overriding that default (we're currently doing that in latest master). I don't think that is correct and we should rely on the default used by the daemon internally.

src/ipfsd-daemon.js
dryajov: good catch

src/ipfsd-daemon.js
dryajov: this is done for backwards compatibility, that is how we used to do it previous versions, plus it helps with chaining in `async/waterfall` type constructs - but not a real contention point for me, if noone objects, I'll remove.

diasdavid: please remove.

diasdavid:
What's missing in this PR? @dryajov ?

src/ipfsd-daemon.js
dryajov: `init` also accepts a keysize

dryajov:
We had this discussion in already, but I wanted to document why I believe that the approach of having ENV variables taking precedence over method arguments is wrong. First, to clarify one more time what I'm advocating for:

(top to bottom, top entry has highest precedence)
- command line options/method arguments
- env variables
- default values

I don't want to talk specifically about the keysize, but in general about "environment setup" and precedence overall. I believe this is a subtle but very important issue, that can hinder flexibility and behave in a counter intuitive way for the user (_this last one is subjective_)

I did some research and while there don't seem to be any official standard for this, it is a widely used convention, a pretty representative one is `gcc`, from the official gcc docs:

https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html (second paragraph)

> Note that you can also specify places to search using options such as -B, -I and -L (see Directory Options). These take precedence over places specified using environment variables, which in turn take precedence over those specified by the configuration of GCC. See Controlling the Compilation Driver gcc in GNU Compiler Collection (GCC) Internals.

The second sentence lays out the rules of precedence that I listed above.

## Why I think this is important.

Doing it this way allows us to have control over the application defaults, but don't prevent the user from overriding those when it is required. A few use cases that come to mind. 

- We want to override the default repo location (`~/.ipfs`) for `disposable: false` daemons, we can set the `IPFS_PATH` variable to an alternative path, but still allow all of our repo specific tests to run with the provided repo to `spawn`.

- We want to set the default `keysize` to something other than 4096 (the default) system wide, we can use the `IPFS_KEYSIZE` variable, but still have specific tests that require running with different keysizes to continue doing so.

There are probably other cases that we would want to handle through environment variables that we haven't yet come across. 

Another issue with having ENV taking precedence is unexpected behavior for the user (again this is subjective, but working with posix utils that follow this convention has made me expect this and so  many other people, IMO). If an ENV var is set to some value, but I am explicitly passing a different value in my application, I expect that value to take precedence over the env - anything else is confusing.

A benefit of going with what i'm proposing here, is that it allows us to enable this across all of the pertinent repos (js-ipfs, ipfs-api, etc...) without modifying any of the tests - just pass `IPFS_KEYSIZE=<size>` to aegir in the package.json as I'm already doing it here.

Sorry for the long post and possibly "beating on a dead horse", and please do let me know where and if I'm missing the point.

@diasdavid let me know if you just want me to make the change in https://github.com/ipfs/js-ipfsd-ctl/pull/203#pullrequestreview-97385617 regardless.

diasdavid:
@dryajov thank you for the very well thought through proposal and literature check. For the sake of moving things along, the reality is that we don't need env vars to solve the _slow tests problem_. Can we take that feature out into a separate PR  with that same proposal you made?

dryajov:
@diasdavid I'll go ahead and make the changes in - https://github.com/ipfs/js-ipfsd-ctl/pull/203#pullrequestreview-97385617

diasdavid:
@dryajov can I get a status update here? Is this PR ready? Have you tested it with the dependents to ensure everything works well? Is there PRs to update the keysize for those repos?

dryajov:
This PR is ready. I'll open up PRs for the remaining repos.

src/endpoint/routes.js
```diff
@@ -96,12 +96,12 @@ module.exports = (server) => {
...
       const payload = request.payload || {}
 
-      nodes[id].init(payload.initOpts, (err, node) => {
+      nodes[id].init(payload.initOpts, (err) => {
         if (err) {
           return reply(boom.badRequest(err))
         }
 
-        reply({ initialized: node.initialized })
+        reply({ initialized: true })
```
diasdavid: Why did this get removed? `nodes[id]` is still available within scope.

src/ipfsd-daemon.js
```diff
@@ -57,6 +58,7 @@ class Daemon {
     this._gatewayAddr = null
     this._started = false
     this.api = null
+    this.bits = process.env.IPFS_KEYSIZE || (this.opts.initOptions ? this.opts.initOptions.bits : null)
```
diasdavid: leftover from moving the env vars to a separate PR.

src/ipfsd-daemon.js
```diff
@@ -111,41 +113,50 @@ class Daemon {
...
    * @param {function (Error, Node)} callback
    * @returns {undefined}
    */
-  init (initOpts, callback) {
+  init (initOptions, callback) {
     if (!callback) {
-      callback = initOpts
-      initOpts = {}
+      callback = initOptions
+      initOptions = {}
```
diasdavid: Does this mean that is mandatory to always pass a initOptions object? There should be a check to see if initOptions is a func.

src/ipfsd-daemon.js
diasdavid: Can you update this comment to refer the WIP PR https://github.com/ipfs/js-ipfs/pull/785 and add a proper TODO to come back here once that is done.

src/ipfsd-in-proc.js
```diff
@@ -32,8 +35,9 @@ class Node {
     this._started = false
     this.initialized = false
     this.api = null
+    this.bits = process.env.IPFS_KEYSIZE || (this.opts.initOptions ? this.opts.initOptions.bits : null)
```
diasdavid: leftovers.

src/ipfsd-in-proc.js
```diff
@@ -113,36 +118,41 @@ class Node {
...
    * @param {function (Error, Node)} callback
    * @returns {undefined}
    */
-  init (initOpts, callback) {
+  init (initOptions, callback) {
     if (!callback) {
-      callback = initOpts
-      initOpts = {}
+      callback = initOptions
+      initOptions = {}
```
diasdavid: Same as above.

test/add-retrieve.spec.js
```diff
@@ -25,7 +25,9 @@ describe('data can be put and fetched', () => {
 
       const f = IPFSFactory.create(dfOpts)
 
-      f.spawn((err, _ipfsd) => {
+      f.spawn({
+        initOptions: { bits: 1024 }
```
diasdavid: Put the initOptions as part of the tests array. This way you can do 512 for js-ipfs and 1024 for go-ipfs

test/api.spec.js
```diff
@@ -110,7 +114,10 @@ describe('ipfsd.api for Daemons', () => {
     it('check if API and Gateway addrs are correct', function (done) {
       this.timeout(30 * 1000)
```
diasdavid: These timeouts are probably no longer needed (or less time) with shorter keysize. Please check all of them.

test/spawn-options.spec.js
```diff
@@ -112,6 +116,9 @@ describe('Spawn options', () => {
       })
 
       describe('spawn from a initialized repo', () => {
+        // TODO: wont work on windows until we get `/shutdown` implemented in js-ipfs
+        if (isWindows) { return }
```
diasdavid: Please use a `this.skip()`

test/start-stop.node.js
```diff
@@ -180,34 +187,45 @@ types.forEach((type) => {
...
             expect(node).to.not.exist()
             done()
           })
-        })
       })
     })
 
     describe('start and stop multiple times', () => {
+      // TODO: wont work on windows until we get `/shutdown` implemented in js-ipfs
+      if (isWindows) { return }
```
diasdavid: Please use this.skip()

test/start-stop.node.js
```diff
@@ -244,7 +264,7 @@ types.forEach((type) => {
         expect(ipfsd).to.exist()
       })
 
-      it('daemon should not be running', (done) => {
+      it('daemon should be running', (done) => {
```
diasdavid: Why this change?

test/start-stop.node.js
dryajov: Skip doesn't work inside `describes` only `it` blocks.

test/spawn-options.spec.js
dryajov: Skip doesn't work inside `describes` only `it` blocks.

test/start-stop.node.js
dryajov: It was misnamed, we are explicitly initing and starting the node.

test/start-stop.node.js
diasdavid: @dryajov can we put that inside every it block then? It signals the viewer that something is not complete.

test/spawn-options.spec.js
diasdavid: https://github.com/ipfs/js-ipfsd-ctl/pull/203#discussion_r169578144

src/ipfsd-daemon.js
dryajov: we're checking for `!callback` which is not consistent with how do it elsewhere (leftover from prev codebase), I'll change it. Good catch.

test/add-retrieve.spec.js
```diff
@@ -11,8 +11,8 @@ const IPFSFactory = require('../src')
 const JSIPFS = require('ipfs')
 
 const tests = [
-  { type: 'go' },
-  { type: 'js' },
+  { type: 'go', bits: 1024 },
+  { type: 'js', bits: 512 },
   { type: 'proc', exec: JSIPFS }
 ]
```
diasdavid: The proc node also should receive the bits.

diasdavid:
Is there any blockers from getting Jenkins to pass?

dryajov:
no, I know what it is, pushing shortly.

diasdavid:
Seems that -- https://github.com/ipfs/js-ipfs/pull/1103#issuecomment-366779893 -- is also captured here:

image

test/add-retrieve.spec.js
```diff
@@ -11,8 +11,8 @@ const IPFSFactory = require('../src')
 const JSIPFS = require('ipfs')
 
 const tests = [
-  { type: 'go' },
-  { type: 'js' },
+  { type: 'go', bits: 1024 },
+  { type: 'js', bits: 512 },
   { type: 'proc', exec: JSIPFS }
```
diasdavid: Don't forget about his one.

dryajov:
Yes, at this point I'm not sure if there is something running on those ports on mac os already, or something in ipfs or elsewhere is holding on to those ports.

package.json
```diff
@@ -78,17 +78,19 @@
     "detect-node": "^2.0.3",
     "hapi": "^16.6.2",
     "hat": "0.0.3",
+    "ipfs-api": "^17.3.0",
```
diasdavid: While you are at it, mind upgrading ipfs-api to latest?
image

dryajov:
my simplistic `time` measurements have yielded ~2 mins :)!

diasdavid:
Last missing issue is: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfsd-ctl/detail/PR-203/17/pipeline/13#step-97-log-272 https://github.com/ipfs/js-ipfsd-ctl/pull/203#issuecomment-367283259. 

Let's handle that one in a separate PR so that we reduce the amount of time spending in testing in the other projects today and we figure out why Jenkins doesn't like how the factory-endpoint terminates.

diasdavid:
Seems that Travis is also not Happy:

image

dryajov:
yep, looking into it now.

dryajov:
PRs to use the new reduced keysizes

- [ ] https://github.com/ipfs/js-ipfs-api/pull/698
- [ ] https://github.com/ipfs/js-ipfs/pull/1232

dryajov:
ok, CIs are gree, I think this is ready?

diasdavid:
@dryajov you are missing the ipfs/interop tests, those daemons can have their keys set to `1024 bits`.

Also, did you test everything locally? Everything 100% passing flawlessly?

dryajov:
Updated interop as well. Looking into why `ipfs-api` is failing.

PR #204, opened 2018-02-16
fix: don't pass on arguments from node executable

### LINKED ISSUE #202: Node inspector doesn't work: address already in use
Hello guys, thank you for your amazing work. 

I use `ipfsd-ctl` for spawning IPFS instance in mocha test in a very similar way as in OrbitDb tests.
It's cool but it doesn't work with node inspector:
```
node_modules/.bin/mocha --inspect test/**/*.js

Debugger listening on ws://127.0.0.1:9229/7c0ad2d6-146b-4ae1-bf2c-7e9873c18304
For help see https://nodejs.org/en/docs/inspector


  Storage
    1) "before all" hook: before


  0 passing (33ms)
  1 failing

  1) Storage
       "before all" hook: before:
     Error: non-zero exit code 12

            while running: /Users/ivanshumkov/Projects/dashevo/dashdrive/node_modules/go-ipfs-dep/go-ipfs/ipfs init -b 2048


            Starting inspector on 127.0.0.1:9229 failed: address already in use

      at Stream.listeners.done.once (node_modules/ipfsd-ctl/src/exec.js:39:11)
      at Stream.f (node_modules/once/once.js:25:25)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:198:12)

```

vmx:
When running a script that involves `ipfsd-ctl` and you pass on
additional parameters it used to pass on those parameters to
the subprocesses it spaws. This lead to issues especially with
`--inspect-brk` as it would spawn another inspector which then
would fail because there is already one running on the same port.

Fixes #202.

### DIFF
--- a/src/utils/run.js
+++ b/src/utils/run.js
@@ -13,5 +13,8 @@ module.exports = (node, args, opts, callback) => {
     executable = process.execPath
   }
 
+  // Don't pass on arguments that were passed into the node executable
+  opts.execArgv = []
+
   return exec(executable, args, opts, callback)
 }


dryajov:
@vmx does it work :)?

LGTM otherwise 👍

vmx:
@dryajov It works for me. I currently use it locally. Though I'd like to have things pass at least once on all platforms as two tests fail locally for me (due to other reasons than they fail n CI).

dryajov:
Awesome, that's an essential feature to have 👍

PR #205, opened 2018-02-17
Graceful stop

richardschneider:
- Use `api/v0/shutdown` to kill a daemon process
- Get tests working on windows

See https://github.com/ipfs/js-ipfs/issues/1192 for background details.

Note that this is blocking on https://github.com/ipfs/js-ipfs/issues/1224.  When fixed, change the `package.json` to use the latest ipfs.


richardschneider:
@diasdavid any comments?

package.json
```diff
@@ -100,7 +100,7 @@
     "detect-port": "^1.2.2",
     "dirty-chai": "^2.0.1",
     "go-ipfs-dep": "0.4.13",
-    "ipfs": "^0.27.5",
+    "ipfs": "github:ipfs/js-ipfs",
```
diasdavid: Let's old this PR until we have a new release of js-ipfs (it can happen soon).

dryajov: It's here, can't wait to get this in :)

screen shot 2018-02-21 at 12 38 57 pm

dryajov:
@richardschneider I'm trying to figure out why the tests are failing now, but having been able to do so yet. As for the port issue, we were able to determine that it's not related to `ipfsd-ctl` itself, but rather something running on those ports on mac os - #209.

richardschneider:
@dryajov cheers!

dryajov:
The issue seems to be somehow related to the `version` entry missing from IndexDB on first run. I can't debug it, since running the set of tests in the browser a second time around works fine, it only seems to fail on first run. Here is the error:

screen shot 2018-02-22 at 4 47 55 pm

Skipping the version test allows the rest of the tests to pass.

dryajov:
@diasdavid @VictorBjelkholm any idea of what might be causing this - https://github.com/ipfs/js-ipfsd-ctl/pull/205#issuecomment-367849964?

richardschneider:
@diasdavid Thanks for the ipfs v0.28 release.  This PR is now ready for review and merge.

diasdavid:
@richardschneider all CI is failing. Errors:

image

image

Let's be attentive of everyone's time and request a review or a merge only when the code is fully ready.

JonKrone:
@richardschneider @dryajov

Jenkins and Travis are green ✔️

richardschneider:
@diasdavid Is this now ready for a merge?

src/factory-in-proc.js
```diff
@@ -65,10 +65,14 @@ class FactoryInProc {
...
-    const IPFS = this.exec
-    new IPFS(options).version((err, _version) => {
-      if (err) { callback(err) }
-      callback(null, _version)
+
+    const IPFS = this.exec // otherwise lint barfs
+    const ipfs = new IPFS(options)
+    ipfs.once('ready', () => {
+      ipfs.version((err, _version) => {
+        if (err) { callback(err) }
```
diasdavid: needs a return. Otherwise chaos will happen.

src/ipfsd-daemon.js
```diff
@@ -270,36 +270,40 @@ class Daemon {
...
+   * If the HTTP API is established, then send 'shutdown' command; otherwise,
+   * process.kill(`SIGTERM`) is used.  In either case, if the process
+   * does not exit after 10.5 seconds then a `SIGKILL` is used.
    *
    * @param {function()} callback - Called when the process was killed.
    * @returns {undefined}
    */
   killProcess (callback) {
     // need a local var for the closure, as we clear the var.
+    const self = this
```
diasdavid: Self is not needed.

richardschneider: Disagree, it is needed in `subprocess.once('exit', ...`

`this` should not be used in a lambdaexpression!  See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

src/factory-in-proc.js
richardschneider: Agree, extremely bad code!!!

src/factory-in-proc.js
```diff
@@ -65,10 +65,14 @@ class FactoryInProc {
...
-      if (err) { callback(err) }
-      callback(null, _version)
+
+    const IPFS = this.exec // otherwise lint barfs
+    const ipfs = new IPFS(options)
+    ipfs.once('ready', () => {
+      ipfs.version((err, _version) => {
+        if (err) { return callback(err) }
+        callback(null, _version)
+      })
```
diasdavid: Is this starting the node? I'm sure this is causing repo lock just to check the version.

src/factory-in-proc.js
```diff
@@ -129,15 +133,16 @@ class FactoryInProc {
...
-
-    series([
-      (cb) => options.init
-        ? node.init(cb)
-        : cb(),
-      (cb) => options.start
-        ? node.start(options.args, cb)
-        : cb()
-    ], (err) => callback(err, node))
+    node.exec.once('ready', () => {
```
diasdavid: Something isn't quite right here. the node is created but then what we listen on the ready event is `exec`??

src/factory-in-proc.js
JonKrone: Hey @dryajov, curious about this as [there is a TODO](https://github.com/ipfs/js-ipfsd-ctl/blob/master/src/factory-daemon.js#L66) to fix this on the go/js factory. Did the mentioned solution hit roadblocks originally? Can we make an issue and work on it separate from this PR?

dryajov:
I'm reworking `proc` nodes to propagate the `ready` event. That is the only reliable way of ensuring that the node is fully constructed when interacting with it. Ideally, the user doesn't have to care about it as it would be taken care of by the factory.

richardschneider:
@dryajov Seems that this PR is being re-purposed.  Can we get some closure?

dryajov:
This change is required to get all CIs green. I'm fine with making the change in a different PR but this would depend on it. I'll create a separate PR.

richardschneider:
@dryajov If it's need then please add it here.  I'm just concerned that changes are creeping in here that have nothing to do with stopping.

dryajov:
jenkins is failing because of:

screen shot 2018-03-13 at 12 07 50 am

diasdavid:
@dryajov Travis still fails with a bunch of timeouts

src/factory-in-proc.js
```diff
@@ -65,10 +65,13 @@ class FactoryInProc {
...
-    new IPFS(options).version((err, _version) => {
-      if (err) { callback(err) }
-      callback(null, _version)
+
+    const node = new Node(options)
+    node.once('ready', () => {
+      node.init((err) => {
+        if (err) { return callback(err) }
+        node.version(callback)
+      })
```
diasdavid: It is very odd to me that we start a new instance of IPFS to get the version. This will create a memory leak + it will lock the repo for a consequent node spawn. @dryajov can you expand why to do it this way?

src/ipfsd-in-proc.js
```diff
@@ -68,6 +72,8 @@ class Node {
       EXPERIMENTAL: this.opts.EXPERIMENTAL,
       libp2p: this.opts.libp2p
     })
+
+    this.exec.once('ready', () => this.emit('ready'))
```
diasdavid: Why use a different pattern than the one used for ipfs-daemon (new Node + node.start)?

src/factory-in-proc.js
dryajov: @diasdavid I'm not aware of any other way to get the version - is there a better one you can suggest?

dryajov: btw, the instance is not started, it is simply initialized with a repo, to start it we would have to call `node.start`.

src/ipfsd-in-proc.js
dryajov: I believe not doing that can end up in race conditions - take a look at [maybeOpenRepo](https://github.com/ipfs/js-ipfs/blob/d945fceeb09c1c4289a61edb40209271e8318d25/src/core/boot.js#L67), its called by the IPFS constructore through `boot(self)`, but it is async, and the only way of knowing when it finished is by hooking into the `ready` event.

diasdavid: What I'm weird'ed out is that in `ipfsd-in-proc`, the IPFS instance starts (and therefore does network activity, lock repo, etc) on the constructor, while in `ipfsd-daemon`, the daemon only starts when .start is called. This asymmetry will lead to confusion and debugging issues in the future.

src/factory-in-proc.js
diasdavid: @dryajov are you saying that a repo is created to just get the version of ipfs? Sound like the node.version should work offline/un'init for the info it can provide, namely, the impl version.

If it was absolutely impossible (which it isn't), at least there should be a cleanup step here.

src/ipfsd-in-proc.js
dryajov: ah, I see - we're not actually starting the node since IPFS is constructed with both, `init` and `start` as false, take a look here - https://github.com/ipfs/js-ipfsd-ctl/blob/a61fa50d6834e61465398cd1d54e60ec913358eb/src/ipfsd-in-proc.js#L67...L74. This was done specifically to keep that symmetry with the other types of daemons, the `ready` event is there to make sure we don't trip over anything.

src/factory-in-proc.js
dryajov: Also, I thought about just using `ipfs-repo` directly to read the version, but it turns out we still need to initialize a repo to get the version, since there is no guarantee that the default repo in `~/.jsipfs` corresponds to the version of the IPFS in our `exec` param.

dryajov: Well, it turns out that the version is read by the repo here - https://github.com/ipfs/js-ipfs/blob/494da7f8c5b35f491f22a986ff5e8c456cc0e602/src/core/components/version.js#L14...L24. I think that's an overkill and we should separate the repo version from the ipfs version, that way we don't need a repo. If noone objects I can go ahead and make that change in `IPFS`.

diasdavid: There is a need for a Repo version and IPFS version, that is correct. It should be ok to get the IPFS version without the Repo existing though.

dryajov:
So the issue ended up being browser only. By default, if the repo doesn't exist, the static repo version is used, but in the browser the error returned was not being matched correctly, hence to version was being returned at all. The fix in https://github.com/ipfs/js-ipfs/pull/1262, should address that.

diasdavid:
Thank you @dryajov :) I'll release a patch version of js-ipfs as soon as I have decent Internet again (npm is not friendly of in-flight wifi)

src/ipfsd-in-proc.js
diasdavid: I see. Makes sense then. Thanks for clarifying :)

diasdavid:
Seems that it is all good \o/ 

image

Mac OS fails because Jenkins doesn't remember how to install npm anymore //cc @VictorBjelkholm

Issue #206, opened 2018-02-17
Browser tests failing on MacOS


richardschneider:
What is port 9999? 

See build https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfsd-ctl/detail/graceful-stop/1/pipeline/14

```
Test Browser
listen EADDRINUSE 127.0.0.1:9999
Error: listen EADDRINUSE 127.0.0.1:9999
    at Object._errnoException (util.js:1031:13)
    at _exceptionWithHostPort (util.js:1052:20)
    at Server.setupListenHandle [as _listen2] (net.js:1367:14)
    at listenInCluster (net.js:1408:12)
    at GetAddrInfoReqWrap.doListen [as callback] (net.js:1523:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:102:10)
error Command failed with exit code 1.

info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

richardschneider:
@dryajov Any ideas?  Haven't we seen this before?

dryajov:
I've pinged @VictorBjelkholm about it on IRC - I believe he was to look into it. We can also change the default port for the http server to something more random, but I believe this indicates an issue somewhere else as well - https://github.com/ipfs/js-ipfs/pull/1103#issuecomment-366670887

diasdavid:
<img width="721" alt="image" src="https://user-images.githubusercontent.com/1211152/36532706-5f78d90c-17b9-11e8-9deb-b21b06c2d88c.png">

@dryajov this is a ipfsd-ctl issue, also happening on Jenkins here - https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfsd-ctl/detail/PR-203/19/

**Making it be a port random would only hide the issue, the real problem is that the http server is not closed properly**.

VictorBjelkholm:
@diasdavid I'm working on this right now.

Are you sure about the real problem? I'll make sure it's closed then, but it could also be that the port is used on macOS for other purposes. Real fix would be to make sure port is ephemeral + making sure server is closed IMHO.

diasdavid:
Update: Let's see https://github.com/ipfs/js-ipfsd-ctl/pull/209 :)

VictorBjelkholm:
[...]

PR #209, opened 2018-02-22
fix: try with a different port

diasdavid:


VictorBjelkholm:
:-1: not a proper fix as this might break in the future too. We cannot make assumptions about free ports (but I bet you already knew this). As mentioned elsewhere, I'm working on fixing this right now.

diasdavid:
@VictorBjelkholm I'm testing you hypothesis from https://github.com/ipfs/js-ipfsd-ctl/issues/206#issuecomment-367635620

I claim that it isn't the port but rather the server that is not being closed properly.

diasdavid:
Alright, seems that indeed Mac OS uses port 9999 for something, so the server is being closed properly.

image

👌🏽

VictorBjelkholm:
> Making the port random generated would just hide the problem for now, but it would create issues for users

The fix I'm working on doesn't involve ipfsd-ctl at all, just how ipfs/js-ipfs is using it.

dryajov:
I say we change the port to something else - the one here seems reasonable, `9999` seems too common and super easy to clash with. @diasdavid lets get this merged?

VictorBjelkholm:
Silly but `4737` would mean `IPFS` on a T9 keyboard (and US phonenumbers somehow?). Would be a good port and easy to remember if you used to send sms's a couple of years ago :tada:

diasdavid:
#211

diasdavid:
@VictorBjelkholm that one would be good for the ipfs daemon default listening port :)

PR #211, opened 2018-02-23
fix: try with a different port

diasdavid:
Given that a solution is only being applied temporarily here https://github.com/ipfs/js-ipfs/pull/1103/files#r169972169 by @VictorBjelkholm, let's at least have the tests passing with flying colors and create a proper solution in aegir itself, tracked https://github.com/ipfs/aegir/issues/201


diasdavid:
Last issue with one of the Windows runs seems #205 related.

PR #212, opened 2018-02-28
fix: daemon should recognize previously created repos.

JonKrone:
While working on some go/js interop tests I noticed that a daemon spawned on an already-initialized repo doesn't report itself as initialized.

It turned out to be just a typo so I fixed that and added a test.

**Update**

This PR has four fixes related to `proc` node startup:
- `proc` nodes error when reporting their version: [here](https://github.com/ipfs/js-ipfsd-ctl/pull/212/files#diff-5518d51550f18454b1bdcfed570fbbbcR71)
- `proc` nodes spawn before their IPFS instance has finished booting: [here](https://github.com/ipfs/js-ipfsd-ctl/pull/212/files#diff-5518d51550f18454b1bdcfed570fbbbcR138)
- `proc` nodes do not report themselves as initialized: [here](https://github.com/ipfs/js-ipfsd-ctl/pull/212/files#diff-5518d51550f18454b1bdcfed570fbbbcR139)
- daemons spawned remotely do not report themselves as initialized. [here](https://github.com/ipfs/js-ipfsd-ctl/pull/212/files#diff-b80286715c124db882e2120d455d9ba2R83), [here](https://github.com/ipfs/js-ipfsd-ctl/pull/212/files#diff-7e2d577bc42f51f61a1b45fc5e0d27ceR98), [here](https://github.com/ipfs/js-ipfsd-ctl/pull/212/files#diff-458a8fab61bacfc05620ac20f2c3d6edR24), [here](https://github.com/ipfs/js-ipfsd-ctl/pull/212/files#diff-a64c691b43b66870588c26e5ceb75ea7R55)

### DIFF
--- a/src/factory-in-proc.js
+++ b/src/factory-in-proc.js
@@ -130,7 +130,7 @@ class FactoryInProc {
     const node = new Node(options)
 
     series([
-      (cb) => node.exec.once('ready', cb),
+      (cb) => node.once('ready', cb),
       (cb) => node.repo._isInitialized(err => {
         // if err exists, repo failed to find config or the ipfs-repo package
         // version is different than that of the existing repo.
--- a/src/factory-in-proc.js
+++ b/src/factory-in-proc.js
@@ -128,16 +128,28 @@ class FactoryInProc {
     }
 
     const node = new Node(options)
-    node.once('ready', () => {
-      series([
-        (cb) => options.init
-          ? node.init(cb)
-          : cb(),
-        (cb) => options.start
-          ? node.start(options.args, cb)
-          : cb()
-      ], (err) => callback(err, node))
-    })
+
+    series([
+      (cb) => node.exec.once('ready', cb),
+      (cb) => {
+        try {
+          node.repo._isInitialized(() => {
+            node.initialized = true
+            cb()
+          })
+        } catch (err) {
+          // errors if not initialized
+          node.initialized = false
+          cb()
+        }
+      },
+      (cb) => options.init
+        ? node.init(cb)
+        : cb(),
+      (cb) => options.start
+        ? node.start(options.args, cb)
+        : cb()
+    ], (err) => callback(err, node))
   }
 }
 
--- a/src/factory-in-proc.js
+++ b/src/factory-in-proc.js
@@ -131,18 +131,12 @@ class FactoryInProc {
 
     series([
       (cb) => node.exec.once('ready', cb),
-      (cb) => {
-        try {
-          node.repo._isInitialized(() => {
-            node.initialized = true
-            cb()
-          })
-        } catch (err) {
-          // errors if not initialized
-          node.initialized = false
-          cb()
-        }
-      },
+      (cb) => node.repo._isInitialized(err => {
+        // if err exists, repo failed to find config or the ipfs-repo package
+        // version is different than that of the existing repo.
+        node.initialized = !err
+        cb()
+      }),
       (cb) => options.init
         ? node.init(cb)
         : cb(),
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -52,7 +52,7 @@ class Daemon {
     this.disposable = this.opts.disposable
     this.exec = this.opts.exec || process.env.IPFS_EXEC || findIpfsExecutable(this.opts.type, rootPath)
     this.subprocess = null
-    this.initialized = fs.existsSync(path)
+    this.initialized = fs.existsSync(this.path)
     this.clean = true
     this._apiAddr = null
     this._gatewayAddr = null
--- a/src/ipfsd-in-proc.js
+++ b/src/ipfsd-in-proc.js
@@ -154,7 +154,7 @@ class Node extends EventEmitter {
         (cb) => this.getConfig(cb),
         (conf, cb) => this.replaceConfig(defaults({}, this.opts.config, conf), cb)
       ], (err) => {
-        if (err) { return callback }
+        if (err) { return callback(err) }
         self.clean = false
         self.initialized = true
         return callback()
--- a/src/ipfsd-in-proc.js
+++ b/src/ipfsd-in-proc.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const fs = require('fs')
 const multiaddr = require('multiaddr')
 const defaultsDeep = require('lodash.defaultsdeep')
 const createRepo = require('./utils/repo/create-nodejs')
@@ -33,11 +34,11 @@ class Node extends EventEmitter {
     this.path = this.opts.repoPath
     this.repo = createRepo(this.path)
     this.disposable = this.opts.disposable
+    this.initialized = fs.existsSync(this.path)
     this.clean = true
     this._apiAddr = null
     this._gatewayAddr = null
     this._started = false
-    this.initialized = false
     this.api = null
     this.bits = this.opts.initOptions ? this.opts.initOptions.bits : null
 
--- a/src/ipfsd-in-proc.js
+++ b/src/ipfsd-in-proc.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const fs = require('fs')
 const multiaddr = require('multiaddr')
 const defaultsDeep = require('lodash.defaultsdeep')
 const createRepo = require('./utils/repo/create-nodejs')
@@ -34,7 +33,6 @@ class Node extends EventEmitter {
     this.path = this.opts.repoPath
     this.repo = createRepo(this.path)
     this.disposable = this.opts.disposable
-    this.initialized = fs.existsSync(this.path)
     this.clean = true
     this._apiAddr = null
     this._gatewayAddr = null
--- a/test/spawn-options.spec.js
+++ b/test/spawn-options.spec.js
@@ -138,7 +138,7 @@ describe('Spawn options', function () {
             expect(_ipfsd).to.exist()
             expect(_ipfsd.api).to.not.exist()
 
-            // proc nodes do not reuse initialized repos
+            // proc nodes don't reuse initialized repos
             if (fOpts.type !== 'proc') {
               expect(_ipfsd.initialized).to.eql(true)
             }
--- a/test/spawn-options.spec.js
+++ b/test/spawn-options.spec.js
@@ -114,7 +114,7 @@ describe('Spawn options', function () {
         })
       })
 
-      describe('spawn from a initialized repo', () => {
+      describe('spawn from an initialized repo', () => {
         // TODO: figure out why `proc` IPFS refuses
         // to start with a provided repo
         // `Error: Not able to start from state: uninitalized`
--- a/test/spawn-options.spec.js
+++ b/test/spawn-options.spec.js
@@ -133,11 +133,11 @@ describe('Spawn options', function () {
           }
 
           f.spawn(options, (err, _ipfsd) => {
-            ipfsd = _ipfsd
             expect(err).to.not.exist()
             expect(_ipfsd).to.exist()
             expect(_ipfsd.api).to.not.exist()
 
+            ipfsd = _ipfsd
             // TODO proc nodes don't reuse initialized repos
             // but they should: https://github.com/ipfs/js-ipfsd-ctl/issues/214
             if (fOpts.type !== 'proc') {
--- a/test/spawn-options.spec.js
+++ b/test/spawn-options.spec.js
@@ -133,6 +133,7 @@ describe('Spawn options', function () {
           }
 
           f.spawn(options, (err, _ipfsd) => {
+            ipfsd = _ipfsd
             expect(err).to.not.exist()
             expect(_ipfsd).to.exist()
             expect(_ipfsd.api).to.not.exist()
@@ -142,7 +143,6 @@ describe('Spawn options', function () {
               expect(_ipfsd.initialized).to.eql(true)
             }
 
-            ipfsd = _ipfsd
             done()
           })
         })
--- a/test/spawn-options.spec.js
+++ b/test/spawn-options.spec.js
@@ -79,6 +79,7 @@ describe('Spawn options', function () {
             expect(err).to.not.exist()
             expect(_ipfsd).to.exist()
             expect(_ipfsd.api).to.not.exist()
+            expect(_ipfsd.initialized).to.eql(false)
 
             ipfsd = _ipfsd
             repoPath = _ipfsd.repoPath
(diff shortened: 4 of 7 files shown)


test/spawn-options.spec.js
```diff
@@ -141,6 +141,11 @@ describe('Spawn options', () => {
             expect(_ipfsd).to.exist()
             expect(_ipfsd.api).to.not.exist()
 
+            // proc nodes do not reuse initialized repos
```
dryajov: I think there is no reason why we can't do this in 'proc' nodes as well.

JonKrone: Yeah, we even skip this whole describe block for `proc` nodes just above on line 122. My thought was that proc nodes always set their `initialized` to false ([here](https://github.com/ipfs/js-ipfsd-ctl/blob/master/src/ipfsd-in-proc.js#L36)) so once line 122 gets resolved, this would fail for those nodes.

JonKrone:
@dryajov Travis is failing on browser tests for my changes. There may be an obvious answer for this, but: how are browser tests being run? When I run `test:browser`, no tests run and the output is just an error (below).
```
0 passing (1m)

Error: Some tests are failing
```

As I understand, aegir looks for a `test/browser.js` file, though this project does not have one.
There are clearly browser tests being run on CI so I'm a little confused. I've checked travis config but no reference to anything but `npm run test`

JonKrone:
Ah! Figured it out. aegir fallback on all all tests with `.spec.js`. Turns out my problem was not having Chrome installed! I had a development version of Chrome but it didn't recognize that.

JonKrone:
@dryajov I've looked into why the browser tests are failing and have found something but I don't think it's intended. The daemon that the browser's DaemonClient is communicating with is `initialized` but the browser's DaemonClient doesn't recognize that. If you try to call `init` on the client's daemon, it errors with `ipfs configuration file already exists`

The flow is, I think:

1) browser creates factory, `remote: true` so use `FactoryClient`
2) browser spawns daemon from factory with a `repoPath` that is already initialized
   - browser sends request to `/spawn` endpoint
   - ipfsd-ctl spawns a daemon from `FactoryDaemon`
      - **This spawned daemon has `initialized === true`**
   - ipfsd-ctl responds with the {id, api} of the spawned daemon
   - browser creates a `DaemonClient` with {id, api}
      - **This `DaemonClient` has `initialized === false`**
 3) `DaemonClient` is the ipfsd instance with an `api` we can send commands through.
    - calling `ipfsd.init` fails

Should we instead carry the `initialized` through the `/spawn` api and feed it to the DaemonClient?

JonKrone:
I've just pushed changes for the above problem. Happy to change or revert it if you want me to.

There's is a race condition I was running into during browser tests for [`ipfsd.start`](https://github.com/ipfs/js-ipfsd-ctl/pull/212/files#diff-99efe22f49ec355bf23290dabfc3470eR154) on an initialized repo with a go daemon. I am working on a Windows machine and it may be something that is resolved with the `/shutdown` work. The daemon failed to start on top of a repo because it was still locked.

JonKrone:
**Current status**: CI inconsistent due to race conditions but Jenkins and Travis seem to indicate that things are okay.

**Jenkins** passed Windows 9.2 but failed 8.9 with a daemon.stop race condition (shutdown related).

**Travis** passed Node 6 and Node 8 tests passed but coverage failed. A similar thing happened with [previous CI](https://travis-ci.org/ipfs/js-ipfsd-ctl/jobs/347433462): Travis failed during coverage (but on different tests than this time).

Ready for review

dryajov:
Pity that we don't get CircleCi builds - I'm guessing its because this is a PR from a fork. The existing CI issues indeed look "normal" - windows builds fail because of lack of #205 (as you mention).

test/spawn-options.spec.js
```diff
@@ -137,11 +137,16 @@ describe('Spawn options', () => {
           }
 
           f.spawn(options, (err, _ipfsd) => {
+            ipfsd = _ipfsd
```
dryajov: I believe this will fail lint - can you run `npm run lint` on the repo and make sure everything passes please?

test/spawn-options.spec.js
```diff
@@ -137,11 +137,16 @@ describe('Spawn options', () => {
...
             expect(err).to.not.exist()
             expect(_ipfsd).to.exist()
             expect(_ipfsd.api).to.not.exist()
 
-            ipfsd = _ipfsd
+            // proc nodes don't reuse initialized repos
+            if (fOpts.type !== 'proc') {
```
dryajov: We should add this functionality to `proc` nodes - there is no reason they can't do that. But I'm fine if its in another PR. Could you file an issue to add this please?

src/ipfsd-daemon.js
```diff
@@ -52,7 +52,7 @@ class Daemon {
     this.disposable = this.opts.disposable
     this.exec = this.opts.exec || process.env.IPFS_EXEC || findIpfsExecutable(this.opts.type, rootPath)
     this.subprocess = null
-    this.initialized = fs.existsSync(path)
+    this.initialized = fs.existsSync(this.path)
```
dryajov: 👍

test/spawn-options.spec.js
JonKrone: Ahh cool. I really haven't even looked at in-proc nodes but this is an opportunity. I've filed an issue and am happy to work on it later. https://github.com/ipfs/js-ipfsd-ctl/issues/214

test/spawn-options.spec.js
JonKrone: I have and do not see any failures. Were you expecting it to fail because it's assigning a global? ipfsd is defined above, on line 124 so here we are just setting it.

I moved this assignment from below because if any of the below `expect`s failed, it would inadvertently cause the next `it`s to fail.

dryajov: Usually eslint will complain if the first line in the callback is not handling the error, but perhaps its OK since it happens right after `ipfsd  = _ipfsd`. I would still move the assignment below the `expects`s since if there was an error `_ipfsd` would be `null/empty/invalid`.

JonKrone:
@dryajov 

I have added the fix for in-proc nodes to this PR: https://github.com/ipfs/js-ipfsd-ctl/issues/214. It ended up being a two-liner:

- [in-proc nodes wait](https://github.com/ipfs/js-ipfsd-ctl/pull/212/files#diff-5518d51550f18454b1bdcfed570fbbbcR134) for their IPFS to be 'ready' during spawning
- [initialized state now based on repoPath's existence](https://github.com/ipfs/js-ipfsd-ctl/pull/212/files#diff-d06a3c390787c7b35d2ab649265df1bbR33), same impl as regular js-ipfs daemons.

I think that they are in-line with the original intent of this PR. If we'd prefer to separate them, I am happy to revert.

src/factory-in-proc.js
```diff
@@ -131,6 +131,7 @@ class FactoryInProc {
     const node = new Node(options)
 
     series([
+      (cb) => node.exec.on('ready', cb),
```
dryajov: please use `once` - this event might be triggered several times.

src/ipfsd-in-proc.js
```diff
@@ -29,11 +30,11 @@ class Node {
     this.path = this.opts.repoPath
     this.repo = createRepo(this.path)
     this.disposable = this.opts.disposable
+    this.initialized = fs.existsSync(this.path)
```
dryajov: this might fail in the browser, since it wouldn't be a path... Maybe we can use the `ipfs-repo` modules here to verify it's a repo - tho I've little experience with it.

JonKrone: Whew! Took me a minute to realize we use package.json to define aliases for some files on browsers. That threw me for a loop.

There's a good possibility from `ipfs-repo`: `_isInitialized(callback)`. I'm about to push something, just running all tests. Lemme know what you think.

src/factory-in-proc.js
```diff
@@ -66,9 +66,13 @@ class FactoryInProc {
...
     }
     const IPFS = this.exec
-    new IPFS(options).version((err, _version) => {
-      if (err) { callback(err) }
-      callback(null, _version)
+    const node = new IPFS(options)
+
+    node.once('ready', () => {
+      node.version((err, _version) => {
+        if (err) { callback(err) }
```
diasdavid: needs return.

src/factory-in-proc.js
```diff
@@ -66,9 +66,13 @@ class FactoryInProc {
...
+        callback(null, _version)
+      })
```
diasdavid: Same comment as https://github.com/ipfs/js-ipfsd-ctl/pull/205#discussion_r173673651

src/factory-in-proc.js
```diff
@@ -131,6 +135,19 @@ class FactoryInProc {
     const node = new Node(options)
 
     series([
+      (cb) => node.exec.once('ready', cb),
```
diasdavid: same comment as https://github.com/ipfs/js-ipfsd-ctl/pull/205#discussion_r173673672

src/factory-in-proc.js
```diff
@@ -131,6 +135,19 @@ class FactoryInProc {
...
+      (cb) => {
+        try {
+          node.repo._isInitialized(() => {
+            node.initialized = true
+            cb()
```
diasdavid: Do not call callbacks within try blocks, it results on any error in the future that gets thrown (in any part of the code) to get caught by this try block

src/factory-in-proc.js
JonKrone: Ha! Thanks, you really have an eye for these. Perhaps we could write a lint warning for this 😉?

src/factory-in-proc.js
JonKrone: > Something isn't quite right here. the node is created but then what we listen on the ready event is exec??

@diasdavid Can you rephrase? `node.exec` is the IPFS instance being run by the daemon `node`. What's going on is that `new Node()` instantiates the daemon and returns immediately with a still-booting IPFS instance under `exec`. We could ask consumers to wait for their in-proc daemon's IPFS to be ready but that is inconsistent with the go and js factories, which deliver ready instances.

src/factory-in-proc.js
JonKrone: Great! That's good to know

JonKrone:
@dryajov looking at the js daemon, if you pass a custom IPFS exec: `js-factory.spawn({ exec: new IPFS({}) })`, the daemon returned by `factory.spawn` would still be booting. Do we expect users to know about this and to pass a 'ready' IPFS or wait to be ready after spawning? Could be, just curious.

https://github.com/ipfs/js-ipfsd-ctl/blob/master/src/ipfsd-daemon.js#L53
and
https://github.com/ipfs/js-ipfsd-ctl/blob/master/src/factory-daemon.js#L132

dryajov:
`new IPFS()` is not a valid parameter for `exec` - it expects a class/function declaration not instantiation. Its also not a valid parameter for `go` or `js` types. So the behavior is undefined in those cases.

src/factory-in-proc.js
JonKrone: @diasdavid Looks like there already is an ESLint rule for this: https://eslint.org/docs/rules/callback-return

Adding a lint rule to aegir would be a big change across our projects but it might be worth it for this one as it's not about style but functionality and could catch bugs. It might be okay to set it with error severity but we would want to pre-run some repos to check.

We could alternatively phase it in as a warning and then bump to an error in a month or two, only problem is we don't really pay attention to warnings. If it ends up finding a lot of cases we could do a fix chore of key repos during that period.

richardschneider: I say enable the lint rule in aegir as an error and see what happens.

dryajov:
@JonKrone can we get master rebased here. It should resolve the concerns that @diasdavid raised.

JonKrone:
PR is up to date. I'm seeing 3 CI issues, two of which are new to me:

1) macos EADDIRINUSE. being addressed in: https://github.com/ipfs/jenkins/issues/97
2) windows 9.2 seems to have hung:

JonKrone:
PR is up to date. Jenkins failed due to a macos EADDIRINUSE error (has an issue: https://github.com/ipfs/jenkins/issues/97) but everything else is ✅ 

I've since re-run CI and hit a couple of different issues:

1) macos 9.2 failed to download `go-ipfs_v0.4.13_darwin-amd64.tar.gz`. I saw this issue once before, many weeks ago, and I'm not sure if it is a rare intermittent issue or related to something else: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfsd-ctl/detail/PR-212/16/pipeline/14

2) windows 9.2 seemed to hang on `[5/5] Building fresh packages...` of the `yarn --mutex network` command for about 30 minutes. I stopped it, concerned about wasting resources. I've since learned that it may have been waiting for a build elsewhere on the network to complete before running its final step but I'm not sure. Does someone know if it was hanging or just paused? https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfsd-ctl/detail/PR-212/14/pipeline

VictorBjelkholm:
> PR is up to date. Jenkins failed due to a macos EADDIRINUSE error (has an issue: ipfs/jenkins#97) but everything else is white_check_mark

Yeah, we have something weird with aegir itself or the tests it runs that doesn't properly stop resources when it's done and some processes ends up running, so the next run on the same project + worker leads to collisions...

> macos 9.2 failed to download go-ipfs_v0.4.13_darwin-amd64.tar.gz. I saw this issue once before, many weeks ago, and I'm not sure if it is a rare intermittent issue or related to something else: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfsd-ctl/detail/PR-212/16/pipeline/14

There seems to be some DNS flipping happening with dist.ipfs.io currently, which is why this error is happening, so seems intermittent.

> windows 9.2 seemed to hang on [5/5] Building fresh packages... of the yarn --mutex network command for about 30 minutes.

I think windows is just very slow on installing. The `--mutex network` flag does indeed make sure only run one instance of yarn, but currently is locally to the worker, to prevent concurrent installs (neither yarn or npm handles those very good).

30 minutes sounds like an aweful lot though, I'll look into it.

diasdavid:
@JonKrone please start opening PR from branches from the repo itself to make it simpler to collaborate on a WIP PR.

diasdavid:
image

Reading the above, LGTM. Great work @JonKrone :)

PR #215, opened 2018-03-14
keysize is being ignored

richardschneider:
The keysize was not being passed on the command line. 

`args.concat` does not modify the arrays, it creates a new array.

### DIFF
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -136,8 +136,8 @@ class Daemon {
     // in case we decide to change it at
     // the daemon level in the future
     if (bits) {
-      args.concat(['-b', bits])
-      log(`initializing with keysize: ${bits}`)
+      args.push('-b')
+      args.push(bits)
     }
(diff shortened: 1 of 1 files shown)


src/ipfsd-daemon.js
```diff
@@ -136,8 +136,8 @@ class Daemon {
     // in case we decide to change it at
     // the daemon level in the future
     if (bits) {
-      args.concat(['-b', bits])
-      log(`initializing with keysize: ${bits}`)
+      args.push('-b')
+      args.push(bits)
```
dryajov: Can we log the args array here - something like:

 ```
log(`initializing with args: ${args.join(' ')}`)
``` 

That should help expose issues like this.

richardschneider: `exec` does command line logging.  That's how I found the error.

diasdavid: oh wow. We never got a speed improv then? How did we shave 5 mins the last time?

PR #221, opened 2018-03-20
feat: detect running node

dryajov:


src/ipfsd-daemon.js
```diff
@@ -62,6 +62,21 @@ class Daemon {
     this._env = Object.assign({}, process.env, this.opts.env)
   }
 
+  /**
+   *
+   * @private
+   */
+  get runningNodeApi () {
```
JonKrone: In addition to the `@private` notation, can we also underscore the name: `_runningNodeApi`

src/ipfsd-daemon.js
```diff
@@ -62,6 +62,21 @@ class Daemon {
...
+    let api
+    try {
+      api = fs.readFileSync(`${this.repoPath}/api`)
+    } catch (err) {
+      debug(`Unable to open api file: ${err}`)
```
JonKrone: `log` instead of `debug`, yeah?

test/start-stop.node.js
```diff
@@ -188,6 +188,58 @@ tests.forEach((fOpts) => {
...
+          initOptions: { bits: fOpts.bits },
+          repoPath: ipfsd.repoPath,
+          disposable: false
+        }, (err, daemon) => {
+          expect(err).to.not.exist()
+          daemon.start((err, api) => {
+            expect(err).to.not.exist()
+            expect(api).to.exist()
+            expect(ipfsd.apiAddr).to.be.eql(daemon.apiAddr)
+            daemon.stop(done)
```
JonKrone: necessary to `stop`? The `after` hook does it and if we've made it this far through the expects, `ipfsd` == `daemon`.

src/ipfsd-daemon.js
dryajov: I've removed the private annotation, there is no need for it to be private.

test/start-stop.node.js
dryajov: `daemon` refers to the second started node.

JonKrone: ah! So two daemons communicating to the same remote ipfs. Thanks

SgtPooki:
@lidel @achingbrain @hacdias @daviddias  it doesn't seem like this PR is in the latest ipfsd-ctl package?

https://github.com/ipfs/js-ipfsd-ctl/commit/2fcaa7d

https://unpkg.com/ipfsd-ctl@10.0.6/src/ipfsd-daemon.js

I'm trying to resolve https://github.com/ipfs-shipyard/pinning-service-compliance/issues/25 and I'm not making much progress.

lidel:
@SgtPooki  i suspect detecting `$IPFS_PATH/api`  and using user's node instead of spawning one for tests caused test failures due to version mismatch etc.


I believe the current rule of thumb is to avoid reusing user's node, and always  spawning an ephemeral instance just for tests.

You could customize ports before starting the  node (detect busy portsa with something like [portfinder]( https://www.npmjs.com/package/portfinder)), or pass `test` and `disposable` as `true` to get instance on random ports.

PR #222, opened 2018-03-27
feat: upgrade to go-ipfs-0.4.14

diasdavid:


test/start-stop.node.js
```diff
@@ -141,7 +141,7 @@ tests.forEach((fOpts) => {
           ipfsd.start(['--should-not-exist'], (err) => {
             expect(err).to.exist()
             expect(err.message)
-              .to.match(/Unrecognized option 'should-not-exist'/)
+              .to.match(/unknown option "should-not-exist"\n/)
```
diasdavid: @Stebalien This is a breaking change for the CLI. Not a deal breaker.

Stebalien: *sigh*... Yeah, we should have predefined error *codes* but currently we just spit back messages (and occupationally change them).

Issue #226, opened 2018-04-03
Trying to start non-disposable node fails on second attempt


negamaxi:
Electron app. Trying to start ipfs daemon in a renderer process with this piece of code:
```javascript
    const factory = IPFSFactory.create({ type: "go" })
    const startDaemon = (ipfsd) => {
      ipfsd.start(daemonFlags, (err, api) => {
        if (err) {
          reject(err)
        }
        resolve(api)
      })
    }
    factory.spawn({ disposable: false, repoPath: '/my/repo/path/' }, (err, ipfsd) => {
      if (err) {
        reject(err)
      }
      if (ipfsd.initialized) {
        startDaemon(ipfsd)
      } else {
        ipfsd.init((err) => {
          if (err) {
            resolve(err)
          }
          startDaemon(ipfsd)
        })
      }
    })
```
If repo doesn't exists it works as expected: new repo get initialized, daemon starts, I can see it in my processes list and in `subprocess` field of `ipfsd`. But the next time demon does not start, despite the fact that `initialized` and `_started` fields are all true. No errors, `subprocess` field is null. Am I missing something?

dryajov:
Thanks for reporting this - it should indeed work for your use case. This seems like a bug.

travisperson:
Most likely the reason for this is because under `$IPFS_PATH` there exists an `api` file. In this case, ipfsd-ctl will skip setting up the daemon and simply return an `ipfs-api` instance for the address within the file.

We are only allowing 10500ms before we issue a `SIGKILL`. This will not allow the daemon to properly clean up and will leave around the `api` and `repo.lock` files under `$IPFS_PATH`.

This would be my guess to why you are having this issue. We probably should clean up these files ourselves if we ever forcefully kill the daemon.

dryajov:
Thanks @travisperson! That could indeed be the reason. 

@negamaxi could you check if the `api` file is being left behind after ipfsd-ctl shutdowns? 

We did experiment with the `10500` (increased from `~7000` ms AFAIR), way back when we we're adding js-ipfs support and found it to be sufficient, but it could fail on certain envs I'm guessing.

@hacdias Have you experienced anything like this in `ipfs-desktop`?

dryajov:
@negamaxi I was able to reproduce this behavior by terminating the `ipfsd-ctl` process abruptly, which doesn't give `ipfs` a chance to shutdown gracefully. This could happen under a number of scenarios, but a few obvious ones I can think off are:

[...]

PR #228, opened 2018-04-06
feat: add timeout to stop and killProcess

dryajov:
Adding a timeout param to `stop` and `killProcess` as well as increasing the default grace timeout for non disposable nodes proposed in https://github.com/ipfs/js-ipfsd-ctl/issues/226#issuecomment-379041502.

> - disposable repos do what they do now when calling stop
> - non disposable repos have a longer timeout by default
> - modify ipfsd.stop to allo passing a timeout that allows the user controlling the behavior

~~_note: don't merge, tests still need work_~~

_this is ready for review_

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,11 +218,11 @@ Start the daemon.
 
 Stop the daemon.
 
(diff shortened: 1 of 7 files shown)


vasco-santos:
SGTM @dryajov 

Can you just rebase your PR with master?

PR #232, opened 2018-04-19
[WIP] feat: improve default address overrides

dryajov:
Improving `defaultAddrs` functionality. This should allow, overriding default API, Gateway and Swarm values separately.


src/factory-in-proc.js
```diff
@@ -102,10 +102,19 @@ class FactoryInProc {
       ? options.init
       : true
 
-    if (options.disposable) {
-      options.config = defaults({}, options.config, defaultConfig)
-    } else {
-      const nonDisposableConfig = clone(defaultConfig)
+    const config = clone(defaultConfig)
+    if (options.defaultAddrs) {
+      if (options.config && options.config.Addresses) {
```
VictorBjelkholm: Seems this logic is duplicated in `src/factory-daemon.js `

dryajov: yes, we need a base class to compile all the duplicate logic into one place... I'll add it next.

hugomrdias:
@dryajov can we get this on track ?

hugomrdias:

PR #233, opened 2018-04-27
fix: error event propagation

hugomrdias:
When using node event emitter error events should always be handled and propagated, i'm doing a fix on ipfs cli related to this issue https://github.com/ipfs/js-ipfs/issues/1180 and couldn't catch errors coming from ipfsd this PR fixes that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ yarn.lock
 **/*.log
 test/repo-tests*
 
+.vscode
+.eslintrc
 # Logs
 logs
 *.log
--- a/src/factory-in-proc.js
+++ b/src/factory-in-proc.js
@@ -130,6 +130,7 @@ class FactoryInProc {
     }
 
     const node = new Node(options)
+    node.once('error', err => callback(err, node))
 
     series([
       (cb) => node.once('ready', cb),
--- a/src/ipfsd-in-proc.js
+++ b/src/ipfsd-in-proc.js
@@ -73,6 +73,7 @@ class Node extends EventEmitter {
     })
 
     this.exec.once('ready', () => this.emit('ready'))
+    this.exec.once('error', err => this.emit('error', err))
   }
 
   /**
--- a/src/ipfsd-in-proc.js
+++ b/src/ipfsd-in-proc.js
@@ -72,8 +72,8 @@ class Node extends EventEmitter {
       libp2p: this.opts.libp2p
     })
 
-    this.exec.once('ready', () => this.emit('ready'))
     this.exec.once('error', err => this.emit('error', err))
+    this.exec.once('ready', () => this.emit('ready'))
   }
 
   /**


.gitignore
```diff
@@ -5,6 +5,8 @@ yarn.lock
 **/*.log
 test/repo-tests*
 
+.vscode
+.eslintrc
```
dryajov: Is there any reason we need to ignore the `. eslintrc` file if its not being used? Usually aegir takes care of all our configs, linting included.

hugomrdias: editors need to find a eslint config to provide linter messages and fix problems, because the config is hidden inside the `aegir > eslint-config-aegir > .eslintrc` the editor cant find it.
because of this i need to manually add this file in every repo to have a normal ctrl-s workflow.

hugomrdias:
yes tests fail because of ipfs. with ipfs master linked locally everything is fine.

.gitignore
dryajov: Editors will usually allow pointing their linting tools to custom locations, and with many contributors/editors being used I don't think it would be practical to add a custom entry for each to the `.gitignore` file.

I don't know what our current guidelines in regards to `.gitignore` are, maybe @diasdavid or @VictorBjelkholm can provide more context.

hugomrdias: This works for all editors because this is all based on how eslint works, nothing fancy here 😄

diasdavid:
@travisperson wanna review here? Seems that we have to release this one with CI failing.

src/factory-in-proc.js
```diff
@@ -130,6 +130,7 @@ class FactoryInProc {
     }
 
     const node = new Node(options)
+    node.once('error', err => callback(err, node))
```
Mr0grog: If a node emits an errors *after* starting, won’t this result in `spawn`’s callback getting called more than once? (First with a success, then, later, with an error.)

hugomrdias: Shouldn't happen but ur right error should come first. Good catch, thank you

Mr0grog:
That didn’t actually fix the problem! Try this:

```js
const IPFSFactory = require('.')
const factory = IPFSFactory.create({type: 'proc', exec: require('ipfs')})

factory.spawn({disposable:true}, (error, daemon) => {
  // This callback should only ever execute once, but instead, it’s about to happen twice.
  console.log('Daemon spawned!')
  if (error) {
    return console.error('Spawn error:', error)
  }
  
  // Do an operation, just to make sure we're working
  daemon.api.id((error, id) => {
    if (error) {
      console.error('Could not get ID!')
    }
    else {
      console.log('Daemon ID:', id.id)
    }
    
    // Do something to make stopping fail (this is arbitrary, but let's say
    // anything *could* happen to put the daemon in a bad state)
    daemon.exec._repo.close(error => {
      console.log('Repo prematurely closed')
      
      // And finally, stop the daemon
      console.log('Stopping daemon...')
      daemon.stop(error => console.log('Daemon stopped:', error || 'no error'))
    })
  })
})
```

That *should* print:

```
Swarm listening on /ip4/127.0.0.1/tcp/59666/ipfs/QmbDGaByZoomn3NQQjZzwaPWH6ei3ptzWK7a7ECtS35DKL
Daemon spawned!
Daemon ID: QmbDGaByZoomn3NQQjZzwaPWH6ei3ptzWK7a7ECtS35DKL
Repo closed: null
Stopping daemon...
Daemon stopped: no error
```

But instead, you get:

```
Swarm listening on /ip4/127.0.0.1/tcp/59666/ipfs/QmbDGaByZoomn3NQQjZzwaPWH6ei3ptzWK7a7ECtS35DKL
Daemon spawned!
Daemon ID: QmbDGaByZoomn3NQQjZzwaPWH6ei3ptzWK7a7ECtS35DKL
Repo closed: null
Stopping daemon...
Daemon spawned!
Spawn error: Error: repo is already closed
    [...stack trace removed...]
Daemon stopped: Error: repo is already closed
    [...stack trace removed...]
```

Note the repeated `Daemon spawned!` logs.

dryajov:
Thats a great catch @Mr0grog and a pretty tricky scenario. 

I see two options:

1. remove `error` handler and wrap in a try/catch
```js
try {
    const node = new Node(options)

    series([
      (cb) => node.once('ready', cb),
      (cb) => node.repo._isInitialized(err => {
        // if err exists, repo failed to find config or the ipfs-repo package
        // version is different than that of the existing repo.
        node.initialized = !err
        cb()
      }),
      (cb) => options.init
        ? node.init(cb)
        : cb(),
      (cb) => options.start
        ? node.start(options.args, cb)
        : cb()
    ], (err) => callback(err, node))
} catch (err) {
  return callback(err)
}
```
Wrapping the initialization in a try/catch should take care of all possible error scenarios, but... See next case.

2. remove `error` handler once `series` is done:

```js
    const node = new Node(options)
    const errHandler = (err) => callback(err, node) 
    node.once('error', errHandler)

    series([
      (cb) => node.once('ready', cb),
      (cb) => node.repo._isInitialized(err => {
        // if err exists, repo failed to find config or the ipfs-repo package
        // version is different than that of the existing repo.
        node.initialized = !err
        cb()
      }),
      (cb) => options.init
        ? node.init(cb)
        : cb(),
      (cb) => options.start
        ? node.start(options.args, cb)
        : cb()
    ], (err) => {
      node.removeListener('error', errHandler)
      callback(err, node)
     })
```
This would be the recommended approach, (case 1 can be seen as an anti-pattern of sorts), but it seems like there would still be cases (tho, most likely legit bugs in ipfs), where an error is thrown but initialization/startup is not aborted correctly, so spawn would be again called twice.

There might be a better approach, so would love to hear suggestions.

dryajov:
BTW, if not error handler is registered, EventEmitter will handle it as a special case and throw the error, see - https://nodejs.org/api/events.html#events_error_events. I think, this is pretty safe in our case, because we control the code paths and handling the error with a proper try/catch....

Mr0grog:
Yeah, I would agree that approach 2 is probably the right one.

Some other thoughts that are probably out of scope:

- It might be nice if the `new Node(options)` statement was wrapped in a try/catch, so a caller only has to worry about the callback and not *also* wrap `factory.spawn()` in a try/catch themselves. That would be a change from the current behavior, though.

- It might *also* be nice if the `error` handler on `node.exec` that propagates the error up to `node` wasn’t a `once` handler, so users can reliably listen for error events on `node` (since some folks are certainly asking for [a future where a single error does not shut down a node](https://github.com/ipfs/js-ipfs/issues/1325)).

dryajov:
> It might be nice if the new Node(options) statement was wrapped in a try/catch, so a caller only has to worry about the callback and not also wrap factory.spawn() in a try/catch themselves. That would be a change from the current behavior, though.

If this is happening, it should be considered a bug. But things being as they are, and errors handling being a bit of a mix (callbacks, eventemitter and _possibly_ throws in some third party code), it might not be a bad idea, tho callbacks in particular should not affect this code path, since we're `new`ing the in process nodes.

> It might also be nice if the error handler on node.exec that propagates the error up to node wasn’t a once handler, so users can reliably listen for error events on node...

Hmm... Not sure if I agree... Error events, specially in the EventEmitter sense are exceptions. If the exception was not handled by the library/function/method internally, it should terminate the process (albeit in a graceful manner), but I would not try to recover from such errors. So, `error` should be only thrown when recovery is not possible. Another way of looking at it is - using exceptions to control flow is mostly wrong.

> (since some folks are certainly asking for a future where a single error does not shut down a node)

I think if cases where users want to recover from the error, it should not be thrown as an error in the first place,  but handled internally by ipfs and re-thrown as a warning or logged somewhere, etc...

dryajov:
> Hmmm, I’m not sure I follow. new IPFS() is designed to throw. If I did this:

I'm actually agreeing 👍 sorry for not being clear - we should wrap `new Node(...)` with a try/catch. We should also standardize on our error reporting strategy, we seem to rely on `EventEmitter` to propagate errors, we should stick to it - I think a perfect example is the `joi` config validation, some errors get reported through `.emit('error')` (and throw if no handler is registered), the `joi` validation seems to throw exclusively.

> Totally! I’m not suggesting stopping that — I’m suggesting making the same facility available to a user of ipfsd as to a user of js-ipfs — that you can always listen for the error event (and if you don’t listen, it would still throw, as it always has)

I was referring to this earlier point:

> It might also be nice if the error handler on node.exec that propagates the error up to node wasn’t a once handler, so users can reliably listen for error events on node

I don't think we should change the `.once` for an `.on`, which is what I believe you're suggesting, rather threat all errors as fatal and propagate only once, which is what we do already - https://github.com/ipfs/js-ipfsd-ctl/blob/master/src/ipfsd-in-proc.js#L75. But I'm not totally opposed to it, tho I still stand by the point on all errors being fatal 😄 

----

I think we have several actionable from this discussion (feel free to add more)

- [ ] Fix calling `spawn` twice on error, see - https://github.com/ipfs/js-ipfsd-ctl/pull/233#issuecomment-386670466
- [ ] Figure out what our error propagation strategy is when creating js-ipfs nodes and fix all that don't conform.
- [ ] Figure out which errors are fatal and which aren't - in which case they should not be thrown as errors.

PR #238, opened 2018-05-04
fix: Daemon spawning throws error with the new multiaddr version

### LINKED ISSUE #237: Daemon Spawning throws error with the new multiaddr version
`js-ipfs-api` tests are broken using multiaddr@5.0.0, when spawning the daemon.

Basically, the gateway address generated by `js-ipfsd-ctl` has an additional space at the beginning of the string. As a result, an exception is thrown, as implemented in https://github.com/multiformats/js-multiaddr/blob/master/src/index.js#L41.

```
Uncaught Error: multiaddr " /ip4/127.0.0.1/tcp/49586" must start with a "/"
```

vasco-santos:
Closes #237

### DIFF
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -265,7 +265,7 @@ class Daemon {
         output += String(data)
 
         const apiMatch = output.trim().match(/API (?:server|is) listening on[:]? (.*)/)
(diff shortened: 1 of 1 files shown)


vasco-santos:
The CI is currently breaking, as it needs the last updates on `js-ipfs`.

PR #245, opened 2018-05-14
chore: upgrade ipfs-api to v21

alanshaw:
I `npm link`'d to ipfs@master and the tests passed


vasco-santos:
Thanks @alanshaw ! 🙂

PR #247, opened 2018-05-16
fix:process.env.IPFS_TYPE correctly parse process.env.IPFS_TYPE

wraithgar:
fix for https://github.com/ipfs/js-ipfsd-ctl/issues/246


dryajov:
Thanks for the PR @wraithgar! But I wonder if we really need this at all? This was a left over from an initial attempt at rewriting ipfsd-ctl that never got removed - we don't even have it documented in the readme. Is there any particular use case you're going for?

dryajov:
> The use case here is running tests for js-ipfs-api That environment variable is the only way to get it to use my local js code, as far as I can tell.

Ah, got you. A simple `npm link` should work in that case. If your changes are in `ipfs` and you want to run it against `ipfs-api` doing an `npm link ipfs` inside `ipfs-api` should be enough.

wraithgar:
I'm talking about running the tests in `js-ipfs-api` itself.

The testing situation there is a little muddy cause it definitely feels like there is a duplication of tests, especially if both `js-ipfs` and `js-ipfs-api` are (mostly) calling the tests from `interface-ipfs-core`.

But the situation remains currently that I am trying to move some of the tests in `js-ipfs-api` to use the tests from `interface-ipfs-core` and without this fix I can't test it.

wraithgar:
After a nice discussion with @dryajov the non-working code is going to be removed.

A small bugfix that would have caused an exception if you didn't pass in opts at all is also fixed here.

vasco-santos:
I agree that the code should be removed!

vasco-santos:
Thanks @wraithgar ! 

@dryajov your review is finished?

PR #248, opened 2018-05-16
chore:test fix exec.node.js tests

wraithgar:
There was a suite of tests that were being bypassed during CI with the assumption that the CI environment was buggy.  Turns out it was the tests.  They were failing locally for me too.

It was not using the dirty-chai so the expect.to.not.exist was failing.
It was also trying to trap SIG* with bin/sh which does not recognize those signals (dropping the SIG prefix works in sh)

This SIG* trapping was what I think was making CI tests fail so I enabled testing in CI again.

I waited to make this PR till jenkins had run the tests and these ones are now passing (I know others aren't and there is another issue to address those)

### DIFF
--- a/test/exec.node.js
+++ b/test/exec.node.js
@@ -2,7 +2,10 @@
 /* eslint-env mocha */
 'use strict'
 
-const expect = require('chai').expect
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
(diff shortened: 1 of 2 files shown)


diasdavid:
The only test failing is the commit lint
image

I'm squashing the commit and changing the commit message to fix that.

PR #258, opened 2018-05-21
chore: use tildes for pre-v1 deps

achingbrain:
> Our rule is: Use ~ for everything below 1.0.0 and ^ for everything above 1.0.0. If you find a package.json that is not following this rule, please submit a PR.

https://github.com/ipfs/community/blob/master/js-code-guidelines.md#dependency-versions


dryajov:
hmm... I wish there was a way to enforce this with the current package managers (perhaps there is?), I think that at least npm, will revert `~` to `^` every time something is installed/updated.

achingbrain:
Maybe we could add something to aegir?

achingbrain:
@dryajov Tada: ipfs/aegir#231

dryajov:
@achingbrain ah! Thats awesome, thanks!

dryajov:
@achingbrain can we rebase to resolve the conflicts?

achingbrain:
@dryajov done!

PR #259, opened 2018-05-21
chore: use require.resolve to find ipfs

achingbrain:
If we use `require.resolve` to find the ipfs executable, we don't have to assume anything about the file system (e.g. that a `node_modules` folder exists).  This is useful if you're running a mono-repo and your dependencies have been hoisted, for example.


achingbrain:
Tangentially related to ipfs/aegir/#217

PR #267, opened 2018-06-13
chore: log daemon stdout and stderr

achingbrain:
I find myself making this change all the time to see logs from the daemon when writing interop tests.

With this change to see output from the daemon you can do:

```sh
$ DEBUG=ipfsd-ctl:daemon:* npm test
```

### DIFF
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -13,6 +13,10 @@ const debug = require('debug')
 const os = require('os')
 const hat = require('hat')
 const log = debug('ipfsd-ctl:daemon')
+const daemonLog = {
+  info: debug('ipfsd-ctl:daemon:stdout'),
+  err: debug('ipfsd-ctl:daemon:stderr')
+}
 
 const safeParse = require('safe-json-parse/callback')
 const safeStringify = require('safe-json-stringify')
@@ -246,12 +250,18 @@ class Daemon {
     let output = ''
     this.subprocess = run(this, args, { env: this.env }, {
       error: (err) => {
+        let line = String(err)
+
         // Only look at the last error
-        const input = String(err)
+        const input = line
           .split('\n')
           .map((l) => l.trim())
           .filter(Boolean)
           .slice(-1)[0] || ''
+
+        if (line) {
+          daemonLog.err(line.trim())
+        }
 
         if (input.match(/(?:daemon is running|Daemon is ready)/)) {
           // we're good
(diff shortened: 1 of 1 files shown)


src/ipfsd-daemon.js
```diff
@@ -246,13 +250,21 @@ class Daemon {
...
+        const input = line
           .split('\n')
           .map((l) => l.trim())
           .filter(Boolean)
           .slice(-1)[0] || ''
 
+        line = line.trim()
+
+        if (line) {
+          daemonLog.err(line)
```
vasco-santos: Do you have a particular reason for doing the string trim before? 

I think that the only way it should be used, is if we can receive error strings composed by empty spaces. If this is not the case, what about having the trim only in the log line?

src/ipfsd-daemon.js
```diff
@@ -246,13 +250,21 @@ class Daemon {
     let output = ''
     this.subprocess = run(this, args, { env: this.env }, {
       error: (err) => {
+        let line = String(err)
```
vasco-santos: The name `line` seems kind of unrelated. What do you think of `errString`, or something more related to the string content?

src/ipfsd-daemon.js
```diff
@@ -263,7 +275,15 @@ class Daemon {
         }
       },
       data: (data) => {
-        output += String(data)
+        let line = String(data)
```
vasco-santos: The name `line` seems kind of unrelated. What do you think of `dataString`, or something more related to the string content?

src/ipfsd-daemon.js
achingbrain: The err string has a newline at the end and `debug` will add a newline to anything it's passed so the trim just removes it.  The var isn't used after this so it was just to not do:

```javascript
if (line.trim()) {
   daemonLog.err(line.trim())
}
```

If you prefer I could just do:

```javascript
if (line) {
   daemonLog.err(line.trim())
}
```

src/ipfsd-daemon.js
achingbrain: `err` is a string with a newline character at the end, hence `line`.

`errString` to me would be better as `errorString` but then I've never been a fan of Hungarian notation so I don't think we should encode the type in variable names and certainly not in a language without types, so just `error`.

...But then by convention we'd expect `error` to be an `Error` object, so it confuses the purpose of this callback.  I'd prefer the options to be:

```javascript
{
  onStdErr: (line) => {},
  onStdOut: (line) => {}
}
```

instead of:

```javascript
{
  error: (err) => {},
  data: (data) => {}
}
```

src/ipfsd-daemon.js
achingbrain: See previous comment.

src/ipfsd-daemon.js
vasco-santos: yes, I would go for your second code snippet.

src/ipfsd-daemon.js
vasco-santos: I see your point. I would suggest changing it to `errorLine` or keep as you have it.

src/ipfsd-daemon.js
achingbrain: Done.

Issue #234, opened 2018-04-29
Program run via `exec` and outputting to stderr assumes failure


VictorBjelkholm:
Currently if you try to run a js-ipfs daemon via js-ipfsd-ctl with `DEBUG` set to `*`, ipfsd-ctl will not work correctly and assume that the js-ipfs daemon outputs errors, since the npm module `debug` outputs to stderr.

Rather than just seeing if there is data on stderr and fail if there is, we should save the data and output it if the exit status is not 0.

Line in question that makes the assumption: https://github.com/ipfs/js-ipfsd-ctl/blob/b781fbb77cc461fe09cb625f8cca037ff5942464/src/utils/exec.js#L60

dryajov:
Thanks @VictorBjelkholm! This sounds reasonable and seems like it would fix our logging issues 👍

achingbrain:
This should be fixed by #270

PR #270, opened 2018-06-19
feat: log stderr from daemon

achingbrain:
* Renames handlers to stderr and stdout to better reflect what they do.
* Hooks up stderr handler to data event of daemon's stderr to log data emitted.
* Uses the callback to handle process errors instead of an extra error handler.
* Specify stderr and stdout handlers in the method options instead of in a separate argument.
* No longer buffers all command output automatically as it can lead to out of memory errors on long running processes, instead the calling code can choose to receive stdout/stderr or not.
* Does not treat any data emitted on stderr as a reason to judge a process as failed when no stderr handler is passed

### DIFF
--- a/test/exec.node.js
+++ b/test/exec.node.js
@@ -10,7 +10,6 @@ const isrunning = require('is-running')
 const cp = require('child_process')
 const path = require('path')
 const exec = require('../src/utils/exec')
-
 const os = require('os')
 
 const isWindows = os.platform() === 'win32'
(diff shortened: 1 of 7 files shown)


test/exec.node.js
```diff
@@ -81,8 +113,7 @@ describe('exec', () => {
     const args = hang.concat(tok)
 
     const p = exec(args[0], args.slice(1), {}, (err) => {
-      // `tail -f /dev/null somerandom` errors out
```
achingbrain: I'm not sure this is correct - on Mac OS at least it complains that `somerandom` doesn't exist but then continues running tailing `/dev/null` and doesn't error out.

achingbrain:
This also fixes #234

test/exec.node.js
vasco-santos: @achingbrain I didn't understand this part. 

As far as I understood, on my Mac OS it seems ok, as well as in the CI for macos.

achingbrain: The comment says the command 'errors out', which I took to mean it exits with a non-zero error code.  When I run it, it prints out a message that `somerandom` doesn't exist but continues running.

vasco-santos: So, `tail -f` outputs appended data as a file grows. In this case, the file does not exist, but the process still hanging.

I tried to add some data to the file, which I tried to tail and nothing happened as the file does not exist when the command was executed.

This way, I think we can remove the comment as you did!

PR #287, opened 2018-08-22
fix: fall back to path

achingbrain:
When developing features on `ipfs`/`jsipfs` concurrently you usually have your dev versions installed on the path.

To alleviate the tedium of having to type:

```console
$ IPFS_JS_EXEC=`which jsipfs` IPFS_GO_EXEC=`which ipfs` npm test
```

...we could just look on the path for an `ipfs`/`jsipfs` to run.

The PR uses the [`which`](https://www.npmjs.com/package/which) module to attempt to find `ipfs`/`jsipfs` on the path if all other methods fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "safe-json-parse": "^4.0.0",
     "safe-json-stringify": "^1.1.0",
     "shutdown": "~0.3.0",
-    "stream-http": "^2.8.1",
+    "stream-http": "^3.0.0",
     "superagent": "^3.8.2",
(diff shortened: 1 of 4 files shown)


hugomrdias:
@achingbrain can you check the failing test pls? so we can merge this

diasdavid:
Tests failing on Windows

image

@achingbrain care to take a look?

achingbrain:
I don't think this is needed any more since we now have much better control over which `ipfs` versions are used.

PR #288, opened 2018-08-22
chore: update deps

achingbrain:
There have been a few new libraries released recently


package.json
```diff
@@ -105,17 +105,17 @@
...
+    "aegir": "^15.1.0",
     "chai": "^4.1.2",
     "cross-env": "^5.1.4",
     "detect-port": "^1.2.2",
     "dirty-chai": "^2.0.1",
     "go-ipfs-dep": "0.4.17",
-    "ipfs": "~0.30.1",
-    "is-running": "^1.0.5",
+    "ipfs": "~0.31.7",
+    "is-running": "^2.1.0",
```
diasdavid: Watch out for `is-running` updates. The version 2 of this module left zombie daemons around, not sure if it is still the case but needs to be heavily tested.

achingbrain: That's interesting - I can't find a relevant issue on their repo about this.  From what I can see [v1](https://github.com/nisaacson/is-running/blob/88ab53464543120849bbddbca025a170932fe456/index.js#L23-L29) and [v2](https://github.com/nisaacson/is-running/blob/master/index.js#L5-L10) of the module both just wrap `process.kill(pid, 0)`.

PR #291, opened 2018-08-28
fix: fix ipfsd.init return value

### LINKED ISSUE #289: ipfsd init start docs and behavior don't match
https://github.com/ipfs/js-ipfsd-ctl/blob/master/README.md#initinitopts-callback
should be `ipfsd.init([initOpts], callback)` and also says init returns ipfsd be it doesn't.

https://github.com/ipfs/js-ipfsd-ctl/blob/master/README.md#ipfsdstartflags-callback
says it returns ipfsAPI be doesn't.

Probably all should return ipfsd.

hugomrdias:
init should return ipfsd instance in the callback but didn't

closes #289

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ Start the daemon.
 
 `flags` - Flags array to be passed to the `ipfs daemon` command.
 
-`callback` is a function with the signature `function(err, ipfsApi)` that receives an instance of `ipfs-api` on success or an instance of `Error` on failure
+`callback` is a function with the signature `function(Error, ipfsApi)` that receives an instance of `Error` on failure or an instance of `ipfs-api` on success. 
 
 
 #### `ipfsd.stop([timeout, callback])`
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ Initialize a repo.
   - `directory` (default IPFS_PATH if defined, or ~/.ipfs for go-ipfs and ~/.jsipfs for js-ipfs) - The location of the repo.
   - `pass` (optional) - The passphrase of the key chain.
 
-`callback` is a function with the signature `function (Error, ipfsd)` where `err` is an Error in case something goes wrong and `ipfsd` is the daemon controller instance.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -215,7 +215,7 @@ Start the daemon.
 
 `flags` - Flags array to be passed to the `ipfs daemon` command.
 
-`callback` is a function with the signature `function(err, ipfsApi)` that receives an instance of `ipfs-api` on success or an instance of `Error` on failure
+`callback` is a function with the signature `function(Error, ipfsApi)` that receives an instance of `Error` on failure or an instance of `ipfs-api` on success.
```
vasco-santos: Why changing the  `err` name to `Error`? I think it is more correct to have `function  (err, ipfsApi)` and it will be also consistent with all other `js-ipfs-_` repos.

alanshaw: The changes you've made in the other files do not return an instance of `ipfs-api`.

vasco-santos: @alanshaw these changes are the docs for `start`. His changes in the other files are for the `init`

alanshaw: Ah sorry!

hugomrdias: Init method already uses Error and documentation.js detects the type and gives better output

hugomrdias: Yeah included also this change because I initially thought start method was also not returning correctly.

hugomrdias: I think start should also return ipfsd instead of API.
ipfsd includes the api and waterfalling spawn,init,start for non disposable gets easier. What you guys think?

vasco-santos: I see `documentation.js` detecting the type as an advantage in fact! Maybe we should create a separate discussion for this usage across all the projects, in order to understand the general impression. With this, we could make a consistent transition. What do you think @alanshaw  @hugomrdias ?

vasco-santos: I agree that it should return the `ipfsd`, as it has the `api`. However, this would probably be a breaking change in a lot of repos...

hugomrdias: yes :/ i'll create an issue about that

hugomrdias:
reverted the readme docs to use `err` and ill do a PR about jsdoc/documentation.js for others to review and see if it makes sense.

PR #300, opened 2018-10-04
[WIP] feat: use ipfs-pregen-ids module

mkg20001:
This enables usage of the external [`ipfs-pregen-ids`](https://github.com/mkg20001/ipfs-pregen-ids) module and removes the `testIds`

Removes:
 - Automatic pregen id usage for disposable nodes

Adds:
 - `pregen` option to `.create()`
 - Automatic pregen id usage for disposable nodes _only_ if ipfs-pregen-ids module enabled

Todo:
 - [ ] docs
 - [ ] tests
 - [ ] _invite @mkg20001 to ipfs/javascript-team so I can move this module here?_ Make ipfs-pregen-ids an offical module?

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,6 +118,7 @@
     "dirty-chai": "^2.0.1",
     "go-ipfs-dep": "0.4.17",
     "ipfs": "~0.32.0",
+    "ipfs-pregen-ids": "~0.1.0",
     "is-running": "^2.1.0",
     "mkdirp": "~0.5.1",
     "proxyquire": "^2.0.1",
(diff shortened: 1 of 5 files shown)


diasdavid:
@mkg20001 will you still drive this PR to the finish line?

mkg20001:
Yes, once I get a review on the current state, because I'm unsure if the way I integrated pregen-ids is good.
Also: Should I move pregen-ids to the IPFS repo like mentioned above?

diasdavid:
@mkg20001 I don't think pregen-ids needs to be a module (that gets a bunch of tests and docs.. it is more maintenance burden), instead, it should just be part of this one.

mkg20001:
@diasdavid The reason for this being a module is that the pregenerated ids are big and not everybody needs them, so they're optional. To use them one simply runs `IPFSD.create(require('ipfs-pregen-ids'))` and that's it.

hugomrdias:
old pr gonna close this, feel free to reopen

PR #303, opened 2018-10-24
fix: pass config to init

hugomrdias:
This PR improves ctl daemon performance passing node config directly to init command.

```
- start-stop tests go and js 49 s > 38s  (go needs the same optimization)
- start-stop tests js only 27s > 18s
- full tests 2m 16s > 1m 50s
```

Needs:
- [x] https://github.com/ipfs/js-ipfs/pull/1662


src/ipfsd-daemon.js
```diff
@@ -162,7 +162,10 @@ class Daemon {
     }
 
     const bits = initOptions.bits || this.bits
-    const args = ['init']
+    const args = [
+      'init',
+      this.opts.type === 'js' && JSON.stringify(this.opts.config)
```
diasdavid: why just JS?

diasdavid:
I've rebased master onto this branch.

src/ipfsd-daemon.js
diasdavid: Can we haz this for Go as well?

hugomrdias: i talked about this in the ipfs pr https://github.com/ipfs/js-ipfs/pull/1662#issue-225379313

diasdavid: @Stebalien can you check this ^^

Stebalien: Go actually takes in an initial config as a file argument (positional). However, it won't *merge* it. Really, I'm not sure how I'd expect merging to work.

We also have a concept of "profiles" for patching our config.

hugomrdias: @Stebalien we added support for passing a json string has a positional arg to the init cmd like this https://github.com/ipfs/js-ipfs/pull/1662/files 
We already had a config options https://github.com/ipfs/js-ipfs#optionsconfig so it was straight forward.
It's just a simple deep merge with the default config.

any chance we can get something similar in go ? we can consider other ways to do it! 

we just want to reduce the number of child processes that ctl spawns

Right now we need to spawn 4 child_processes to start a node in ctl
```
ipfs init
ipfs config show
ipfs config replace
ipfs daemon
```

With this PR we would only need 2
```
ipfs init '{"Addresses":{"Swarm":["/ip4/0.0.0.0/tcp/4003"]}}'
ipfs daemon
```

Final goal is to just do 
```
ipfs daemon --init --config "{}"
```

This will help **a lot** making our tests faster!!

Stebalien: A similar issue has come up with in cluster: https://github.com/ipfs/go-ipfs/issues/6262#issuecomment-486938717 (maybe, I may be misunderstanding the issue).

However, I believe they'll need to pass by-file. Would that work here or will we need two options.

daviddias:
@hugomrdias what's missing to land?

hugomrdias:
this change is gonna land with this https://github.com/ipfs/js-ipfsd-ctl/pull/369 PR

PR #304, opened 2018-10-24
[WIP] docs: add example for non disposable daemons

daviddias:
The non-disposable daemons option is currently not working. It errors silently.

Steps to reproduce:

- 1. clone this branch, npm install, go to examples/local-non-disposable and run npm install for it to link the main module
- 2. run ipfs init and jsipfs-init
- 3. node index.js

Result

image

It should spawn a go and js daemons on top of .ipfs and .jsipfs respectively and run the .id call without a problem.

### DIFF
--- a/src/factory-daemon.js
+++ b/src/factory-daemon.js
@@ -98,8 +98,6 @@ class FactoryDaemon {
 
     if (!options.disposable) {
       const nonDisposableConfig = clone(defaultConfig)
-      options.init = false
-      options.start = false
 
       const defaultRepo = path.join(
         process.env.HOME || process.env.USERPROFILE,
@@ -126,9 +124,14 @@ class FactoryDaemon {
     const node = new Daemon(options)
 
     series([
+
+      // TODO if init fails, check if it was because the
+      // repo was already was inited, if yes, continue
       (cb) => options.init
         ? node.init(options.initOptions, cb)
         : cb(null, node),
+      // TODO if start fails, check if it was because there
+      // is a daemon already running and connect to it instead
       (cb) => options.start
         ? node.start(options.args, cb)
         : cb()
--- a/src/factory-daemon.js
+++ b/src/factory-daemon.js
@@ -124,16 +124,29 @@ class FactoryDaemon {
     const node = new Daemon(options)
 
     series([
-
-      // TODO if init fails, check if it was because the
-      // repo was already was inited, if yes, continue
       (cb) => options.init
-        ? node.init(options.initOptions, cb)
+        ? node.init(options.initOptions, (err) => {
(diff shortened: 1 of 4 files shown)


hugomrdias:
You need to init and start, check the link below
https://github.com/ipfs/js-ipfsd-ctl#disposable-vs-non-disposable-nodes

hugomrdias:
I would think whoever introduce this behaviour, tried to make disposable fully automated and non-disposable the flexible version. You just need to pass `.spawn {init: true, start:true}`

It's just the way it works and is documented right now, we can make an issue about it and change the behaviour.

diasdavid:
`You just need to pass .spawn {init: true, start:true}` This is not working for me either (even if I delete init, given that I already have my repo init'ed)

diasdavid:
Can we treat this PR as the issue? Or you prefer me to open another issue?

src/factory-daemon.js
```diff
@@ -126,9 +124,14 @@ class FactoryDaemon {
...
 
     series([
+
+      // TODO if init fails, check if it was because the
+      // repo was already was inited, if yes, continue
       (cb) => options.init
         ? node.init(options.initOptions, cb)
         : cb(null, node),
+      // TODO if start fails, check if it was because there
+      // is a daemon already running and connect to it instead
```
diasdavid: Just needs a bunch of checks here :)

src/factory-daemon.js
```diff
@@ -98,8 +98,6 @@ class FactoryDaemon {
 
     if (!options.disposable) {
       const nonDisposableConfig = clone(defaultConfig)
-      options.init = false
-      options.start = false
```
diasdavid: this was what was making it not start even if passed the options

src/factory-daemon.js
```diff
@@ -127,10 +125,28 @@ class FactoryDaemon {
...
+          }
+        })
         : cb(null, node),
       (cb) => options.start
-        ? node.start(options.args, cb)
+        ? node.start(options.args, (err) => {
+          if (err) {
+            console.log(err)
+            // TODO if start fails, check if it was because there
+            // is a daemon already running and connect to it instead
```
diasdavid: Would appreciate some brainstorming on this one. Would the best just to move this func out https://github.com/ipfs/js-ipfsd-ctl/blob/master/src/ipfsd-daemon.js#L229-L234 and if the daemon is running, catch the addr from the `api` file in the repo?

@hugomrdias?

hugomrdias:
@diasdavid can you check https://github.com/ipfs/js-ipfsd-ctl/pull/308 related to this.

hugomrdias:

Issue #305, opened 2018-10-24
spawn non-disposable with init: true start: true doesn't work


hugomrdias:
Check this https://github.com/ipfs/js-ipfsd-ctl/pull/304

Issue #276, opened 2018-07-04
TypeError: Cannot read property 'id' of null


vaibhav6233m:
I am trying a simple example to spawn IPFS node inside the electron main process.

I am getting this issue if I try to pass custom options inside f.spawn() method.
If I don't pass any options then it starts the node.

See SS:

![screenshot from 2018-07-04 15-47-09](https://user-images.githubusercontent.com/29979122/42271810-dd1ef526-7fa1-11e8-8a80-d82c7f5a2c9e.png)
![screenshot from 2018-07-04 15-47-29](https://user-images.githubusercontent.com/29979122/42271815-dea52f82-7fa1-11e8-9c5f-190eea47be36.png)

Thanks in advance.

vaibhav6233m:
I came to know that if you pass `disposable: true` then the node starts with the settings provided, but I still need to maintain the custom node repo inside electron app config folder.

diasdavid:
Being fixed at #304.

PR #308, opened 2018-10-29
Fix non disposable daemon init/start and attach to running daemons

hugomrdias:
This PR reworks daemon spawn options and handling of start/init, it also makes sure spawned daemons attach themselves to running daemons on the default/provided repo.

Ref: https://github.com/ipfs/js-ipfsd-ctl/pull/304

Fixes https://github.com/ipfs/js-ipfsd-ctl/issues/305
Fixes https://github.com/ipfs/js-ipfsd-ctl/issues/276

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ server.start((err) => {
 `ipfsd-ctl` can spawn `disposable` and `non-disposable` daemons.
 
 - `disposable`- Creates on a temporary repo which will be optionally initialized and started (the default), as well cleaned up on process exit. Great for tests.
-- `non-disposable` - Non disposable daemons will by default attach to any nodes running on the default or the supplied repo. Requires the user to initialize and start the node, as well as stop and cleanup after wards. Additionally, a non-disposable will allow you to pass a custom repo using the `repoPath` option, if the `repoPath` is not defined, it will use the default repo for the node type (`$HOME/.ipfs` or `$HOME/.jsipfs`). The `repoPath` parameter is ignored for disposable nodes, as there is a risk of deleting a live repo.
+- `non-disposable` - Non disposable daemons will by default attach to any nodes running on the default or the supplied repo. Requires the user to initialize and start the node, as well as stop and cleanup afterwards. Additionally, a non-disposable will allow you to pass a custom repo using the `repoPath` option, if the `repoPath` is not defined, it will use the default repo for the node type (`$HOME/.ipfs` or `$HOME/.jsipfs`). The `repoPath` parameter is ignored for disposable nodes, as there is a risk of deleting a live repo.
 
 ## Batteries not included. Bring your own IPFS executable.
 
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 [![Dependency Status](https://david-dm.org/ipfs/js-ipfsd-ctl.svg?style=flat-square)](https://david-dm.org/ipfs/js-ipfsd-ctl)
 [![js-standard-style](https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat-square)](https://github.com/feross/standard)
 [![Bundle Size](https://flat.badgen.net/bundlephobia/minzip/ipfsd-ctl)](https://bundlephobia.com/result?p=ipfsd-ctl)
+
 > Spawn IPFS daemons using JavaScript!
 
 ## Lead Maintainer
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ server.start((err) => {
 `ipfsd-ctl` can spawn `disposable` and `non-disposable` daemons.
 
 - `disposable`- Creates on a temporary repo which will be optionally initialized and started (the default), as well cleaned up on process exit. Great for tests.
-- `non-disposable` - Requires the user to initialize and start the node, as well as stop and cleanup after wards. Additionally, a non-disposable will allow you to pass a custom repo using the `repoPath` option, if the `repoPath` is not defined, it will use the default repo for the node type (`$HOME/.ipfs` or `$HOME/.jsipfs`). The `repoPath` parameter is ignored for disposable nodes, as there is a risk of deleting a live repo.
+- `non-disposable` - Non disposable daemons will by default attach to any nodes running on the default or the supplied repo. Requires the user to initialize and start the node, as well as stop and cleanup after wards. Additionally, a non-disposable will allow you to pass a custom repo using the `repoPath` option, if the `repoPath` is not defined, it will use the default repo for the node type (`$HOME/.ipfs` or `$HOME/.jsipfs`). The `repoPath` parameter is ignored for disposable nodes, as there is a risk of deleting a live repo.
 
 ## Batteries not included. Bring your own IPFS executable.
 
--- a/src/factory-daemon.js
+++ b/src/factory-daemon.js
@@ -1,13 +1,11 @@
 'use strict'
 
-const defaultsDeep = require('lodash.defaultsdeep')
-const clone = require('lodash.clone')
 const series = require('async/series')
+const merge = require('merge-options')
 const path = require('path')
 const tmpDir = require('./utils/tmp-dir')
 const Daemon = require('./ipfsd-daemon')
 const defaultConfig = require('./defaults/config.json')
-const defaultOptions = require('./defaults/options.json')
 
 /** @ignore @typedef {import("./index").SpawnOptions} SpawnOptions */
 
@@ -23,7 +21,7 @@ class FactoryDaemon {
     if (options && options.type === 'proc') {
       throw new Error('This Factory does not know how to spawn in proc nodes')
     }
-    this.options = Object.assign({}, { type: 'go' }, options)
+    this.options = Object.assign({ type: 'go' }, options)
   }
 
   /**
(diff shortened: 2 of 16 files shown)


README.md
```diff
@@ -84,7 +84,7 @@ server.start((err) => {
 `ipfsd-ctl` can spawn `disposable` and `non-disposable` daemons.
 
 - `disposable`- Creates on a temporary repo which will be optionally initialized and started (the default), as well cleaned up on process exit. Great for tests.
-- `non-disposable` - Requires the user to initialize and start the node, as well as stop and cleanup after wards. Additionally, a non-disposable will allow you to pass a custom repo using the `repoPath` option, if the `repoPath` is not defined, it will use the default repo for the node type (`$HOME/.ipfs` or `$HOME/.jsipfs`). The `repoPath` parameter is ignored for disposable nodes, as there is a risk of deleting a live repo.
+- `non-disposable` - Non disposable daemons will by default attach to any nodes running on the default or the supplied repo. Requires the user to initialize and start the node, as well as stop and cleanup after wards. Additionally, a non-disposable will allow you to pass a custom repo using the `repoPath` option, if the `repoPath` is not defined, it will use the default repo for the node type (`$HOME/.ipfs` or `$HOME/.jsipfs`). The `repoPath` parameter is ignored for disposable nodes, as there is a risk of deleting a live repo.
```
diasdavid: "Requires the user to initialize and start the node, as well as stop and cleanup after wards"

This is incorrect. It should spawn if there is not already one.

hugomrdias:
@diasdavid this was already rebased with master I would prefer to keep both PR separate. But if you want to merge them, we should rebase yours with master, merge this one with yours, make sure the example work and then merge it back to this one with fast-foward.

README.md
hugomrdias: With the current changes this PR isn't a breaking change. Because the previous behavior was basically a bug and now we can just pass start: true and it starts. 
I would like this to be explict, only start if the user tells it to start either by option or manually calling the start method.

I can re-work this a bit to explain this better. 

To auto start by default we should open a new issue and discuss the breaking change.

diasdavid: When it is `disposable` it autostarts, it is confusing that `non-disposable` doesn't autostart.

diasdavid:
#304 has now master rebased onto it.

README.md
hugomrdias: I have two points to keep bikeshedding this, sorry :/
- it's out of this PR scope which primarily fixes a bug according to the current behaviour
- making it auto start is a breaking change lots of repos that rely on the current behaviour will need to change code/tests to explicitly pass `init: false, start: false`

I would really prefer to talk/change this on a different issue/pr.

diasdavid: Valid points

diasdavid:
image

diasdavid:
Fails a bunch of tests on the browser with the same error:

```
  1) non-disposable go daemon
       should attach to initialized and running node:
     Error: Error: Request has been terminated
Possible causes: the network is offline, Origin is not allowed by Access-Control-Allow-Origin, the page is being unloaded, etc.
    at request.post.send.end (webpack://ipfsdctl/src/factory-client.js:100 <- node_modules/aegir/src/config/karma-entry.js:344964:27)
    at Request.IpfsdCtl../node_modules/superagent/lib/client.js.Request.callback (webpack://ipfsdctl/node_modules/superagent/lib/client.js:611 <- node_modules/aegir/src/config/karma-entry.js:302263:3)
    at Request.IpfsdCtl../node_modules/superagent/lib/client.js.Request.crossDomainError (webpack://ipfsdctl/node_modules/superagent/lib/client.js:628 <- node_modules/aegir/src/config/karma-entry.js:302280:8)
    at XMLHttpRequest.xhr.onreadystatechange (webpack://ipfsdctl/node_modules/superagent/lib/client.js:703 <- node_modules/aegir/src/config/karma-entry.js:302355:19)
```

Tests failing are:

```
  non-disposable go daemon
    ✓ should fail when passing initOptions to a initialized repo (3ms)
    1) should attach to initialized and running node
    2) should attach to running node with manual start
    3) should not init and start
    4) should init and start
    5) should only init
    6) should only init manualy

  non-disposable js daemon
    ✓ should fail when passing initOptions to a initialized repo (2ms)
    7) should attach to initialized and running node
    8) should attach to running node with manual start
    9) should not init and start
    10) should init and start
    11) should only init
    12) should only init manualy
```

hugomrdias:
it works now 🔥 🔥 
sorry for the big PR but the behaviour between the 3 impls was very different, on the bright side all 3 impls support non-disposable and auto attaching in node and browser.

Pls review again.

src/factory-daemon.js
```diff
@@ -76,63 +72,34 @@ class FactoryDaemon {
...
-      options.start = false
-
-      const defaultRepo = path.join(
-        process.env.HOME || process.env.USERPROFILE,
-        options.isJs
-          ? '.jsipfs'
-          : '.ipfs'
-      )
-
-      options.repoPath = options.repoPath ||
```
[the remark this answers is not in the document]
vasco-santos: I see that the default is defined in the daemon, while before was set twice! looks good then

src/ipfsd-in-proc.js
```diff
@@ -73,29 +75,23 @@ class InProc extends EventEmitter {
...
-  /**
-   * Get the address of connected IPFS HTTP Gateway.
-   *
-   * @member {Multiaddr}
-   */
-  get gatewayAddr () {
-    return this._gatewayAddr
+  setGateway (addr) {
+    this.gatewayAddr = multiaddr(addr)
+    // TODO find out why we set this
```
vasco-santos: did you have any conclusion on it? The `in-proc` node should not need this afaik

src/ipfsd-client.js
```diff
@@ -4,108 +4,88 @@ const request = require('superagent')
...
-   * @param {string} [initOpts.directory=IPFS_PATH] - The location of the repo.
-   * @param {function (Error, Node)} cb
-   * @returns {undefined}
-   */
-  init (initOpts, cb) {
-    if (typeof initOpts === 'function') {
-      cb = initOpts
-      initOpts = {}
+    if (this.initialized && initOptions) {
+      return callback(new Error(`Repo already initialized can't use different options, ${JSON.stringify(initOptions)}`))
```
vasco-santos: We are going in a way of having all the errors using [err-code](https://www.npmjs.com/package/err-code). I know that this is probably one of the repos where it is not that important to have the codes, but since we are creating new ones, I think we can add it here.

vasco-santos: Same happens in the 2 other daemons.

src/ipfsd-in-proc.js
```diff
@@ -63,8 +59,14 @@ class InProc extends EventEmitter {
...
     })
+  }
 
-    this.exec = new IPFS({
+  setExec (cb) {
+    if (this.api !== null) {
+      return setImmediate(() => cb(null, this))
+    }
+    const IPFS = this.opts.exec
+    this.api = this.exec = new IPFS({
```
vasco-santos: why do we need both if we equal both when running `setExec`, which seems the only place where `this.exec` value is modified.

src/utils/repo/browser.js
```diff
@@ -21,6 +16,14 @@ exports.repoExists = function repoExists (repoPath, cb) {
...
+        .catch(e => cb(null, false))
+    }).catch(e => cb(null, false))
+}
+
+exports.defaultRepo = function (type) {
+  return 'ipfs'
+}
+
+exports.checkForRunningApi = function (path) {
+  return null
```
vasco-santos: what is happening here?

test/non-disposable.spec.js
```diff
@@ -0,0 +1,204 @@
...
+        (path, cb) => df.spawn({
+          initOptions: { bits: fOpts.bits },
+          repoPath: path,
+          disposable: false
+        }, cb)
+      ], (err, ipfsd) => {
+        if (err) {
+          return done(err)
+        }
+
```
vasco-santos: should test initialized to be false here?

src/utils/repo/browser.js
hugomrdias: its just browser version, node version checks the file system for the api file in the browser we don't have fs :)

src/ipfsd-in-proc.js
hugomrdias: i'll get back to this in another PR. i have added this to my todo and will about an issue about it.

src/ipfsd-client.js
hugomrdias: one of my okrs is to add error codes everywhere. i would prefer to handle that in another PR

hugomrdias:
CI only has a commitlint error squash will fix that and 2 Windows random errors.

src/factory-daemon.js
```diff
@@ -23,7 +20,7 @@ class FactoryDaemon {
     if (options && options.type === 'proc') {
       throw new Error('This Factory does not know how to spawn in proc nodes')
     }
-    this.options = Object.assign({}, { type: 'go' }, options)
+    this.options = Object.assign({ type: 'go' }, options)
```
achingbrain: Minor nit: we've required `merge-options` above, why not use it here instead of `Object.assign`?

daviddias:
@hugomrdias what's missing for this PR to land?

daviddias:
@hugomrdias not necessary anymore?

hugomrdias:
no 0.47.0 has all this and more

PR #320, opened 2019-02-06
ci: move to travis

hugomrdias:


.travis.yml
```diff
@@ -0,0 +1,40 @@
...
+
+node_js:
+  - '10'
+
+os:
+  - linux
+  - osx
+  - windows
+
+script: npx nyc -s npm run test:node -- --bail
```
vasco-santos: Do we really want to exit as soon as a test fails? 

I feel this as a good thing for running locally! But, in CI there are advantages and disadvantages and I do not know if this is the best approach.

Sure, it is useful to fail with valid information as soon as possible. However, taking into account that we may want to have a complete test running, to get all the failing tests info, I am not sure if this is the best approach in CI. Opinions?

hugomrdias: the way i think about this is, i run tests locally and they should pass. CI is a safety net and so should fail ASAP so i can realize i missed something fast and fix it.
What it your reasoning around "we may want to have a complete test running, to get all the failing tests info" ? Do you have examples when this was useful ?

PR #321, opened 2019-02-11
feat: support passing profile init option

alanshaw:
refs https://github.com/ipfs/js-ipfs-http-client/pull/942


alanshaw:
Before: 4m8.412s
After: 3m11.664s

PR #353, opened 2019-07-03
chore: refactor to async/await

achingbrain:
Refs: ipfs/js-ipfs#1670


.aegir.js
```diff
@@ -15,8 +15,14 @@ module.exports = {
...
-      post: server.stop.bind(server)
+      pre: (callback) => {
+        server.start()
+          .then(() => callback(), callback)
+      },
+      post: (callback) => {
+        server.stop()
+          .then(() => callback(), callback)
+      }
     }
```
hugomrdias: can you change this to just return the promise, aegir with support for hooks promise is release

achingbrain: Done!

PR #362, opened 2019-08-30
chore: add enable-pubsub flag

vasco-santos:
In the context of [ipfs/js-ipfs#2298#pullrequestreview-280953619](https://github.com/ipfs/js-ipfs/pull/2298#pullrequestreview-280953619), we are moving to `enable-pubsub` instead.


vasco-santos:
The codecov is decreasing now, but we will remove the `enable-pubsub-experiment` after this, and there is no need to duplicate tests for now

PR #366, opened 2019-09-04
chore: remove pubsub flags

vasco-santos:
BREAKING CHANGE: pubsub flags for in-proc daemons will not set values anymore

As discussed, we will remove the pubsub flags from the js daemon and will use the config as the way to disable it (will be enabled by default now!)

Ref:

- [ipfs/go-ipfs#6621](https://github.com/ipfs/go-ipfs/issues/6621)
- [ipfs/js-ipfs#2427](https://github.com/ipfs/js-ipfs/pull/2427)

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "dirty-chai": "^2.0.1",
     "go-ipfs-dep": "~0.4.22",
     "husky": "^3.0.4",
-    "ipfs": "ipfs/js-ipfs#master",
+    "ipfs": "~0.37.1",
     "is-running": "^2.1.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -87,7 +87,7 @@
     "dirty-chai": "^2.0.1",
     "go-ipfs-dep": "~0.4.22",
     "husky": "^3.0.4",
-    "ipfs": "ipfs/js-ipfs#master",
+    "ipfs": "~0.37.1",
```
vasco-santos: We can go back to the older version with this

PR #384, opened 2019-09-28
fix: remove IpfsAPI from in proc node

daviddias:
An in-proc node doesn't expose any HTTP API and therefore has no use for the http client.


hugomrdias:

Issue #374, opened 2019-09-13
Remove or change the ipfs-http-client dependency?


achingbrain:
We bundle `ipfs-http-client` with this module, yet require implementations of `go-ipfs` and `js-ipfs` to either be passed to the factory (for in-proc nodes) or installed on the system for (daemon nodes).

This can lead to the situation where you have an `ipfs-http-client` that is incompatible with the `ipfs` version being used due to deps being out of date or miscellaneous API changes and consequently introduces a lot of release churn just for dep updates.

We can pass an implementation of `ipfs-http-client` in the same way as the `js-ipfs` implementation, but this relies on the developer doing the right thing - reading the docs, passing the arg, etc and gives you extra dependencies you aren't going to use.

We are not the only consumers of this library, people seem to use it as a convenient way to start up a node: https://github.com/ipfs/js-ipfsd-ctl/network/dependents?package_id=UGFja2FnZS0xNDQ1MDU1Mw%3D%3D  Though we did just go from callbacks -> promises without much complaint so people are either not upgrading or don't mind the odd breaking change.

We can:

1. Do nothing
1. Change `ipfs-http-client` to be a peerDependency
1. Remove it entirely and require an implementation, similar to how we require an implementation for an in-proc node

Extracted from ipfs/js-ipfs#2446

achingbrain:
In the case of `1. Do nothing`, it adds a gotcha to the use of this module so we then have to continually explain to everyone that our module does the wrong thing by default and you should really not rely on the bundled `ipfs-http-client` and instead you should specify your own and etc etc etc...

achingbrain:
Here's someone discovering the mismatch on their own and encountering surprise: https://github.com/orbitdb/ipfs-log/pull/266#issuecomment-528622376

Issue #315, opened 2018-12-17
Make ipfsd-ctl as robust as calling ipfs daemon manually


olizilla:
With go-ipfs@0.4.18, `ipfs daemon` will start up cleanly even if there is an `api` file  or  a `repo.lock` file or both in `$IPFS_HOME`, as happens if daemon process is not shutdown gracefully.

`ifpsd-ctl` will fail with an `ECONNREFUSED` if there  is an `api`  file in the repo. This causes issues for apps written to use  ipfsd-ctl. The command line tool works fine but the  app can't start.

 A user can create the  issue by quickly `ctrl+c`-ing the  deamon from the cli, or from  an os level shutdown that doens't allow time for ifpsd-ctl to shutdown.

It would be super helpful to make ipfsd-ctl as robust as running to unclean  repos  as `ipfs daemon` is,  otherwise apps will implement their own hacks, or just be unreliable.

See: 
- https://github.com/ipfs/js-ipfsd-ctl/issues/226#issuecomment-379041502
- https://github.com/ipfs-shipyard/ipfs-desktop/pull/722#issuecomment-444139824

achingbrain:
Being able to override lock files would solve ipfs/js-ipfs#522 too.

I thought I'd opened an issue somewhere proposing something like:

```console
$ jsipfs daemon
The repo at ~/.jsipfs is locked, type `yolo` to contine or CTRL+C to abort
(y)olo >
```

..but now I can't find it.  Might have just been in my head.

Issue #217, opened 2018-03-15
Add option to init ipfs node with go-ipfs defaults


olizilla:
IPFS Desktop uses `js-ipfsd-ctl` to start a `go-ipfs` node. If the user doesn't have an existing repo, their node config is initialised with the defaults chosen `js-ipfsd-ctl` rather than those that the user would get if they ran `go-ipfs` manually, which caused us an issue: https://github.com/ipfs-shipyard/ipfs-desktop/issues/617

We can manually copy the `go-ipfs` defaults into the IPFS Desktop app for now, but it would really help to have a way to ask the daemon factory to provide the "regular" defaults.

Issue #208, opened 2018-02-20
In proc node config not applied


alanshaw:
I'm trying to enable `EXPERIMENTAL.dht` for my in-proc node:

```js
df.spawn({ config: { EXPERIMENTAL: { dht: true } } }, /* ... */)
```

The DHT experiment is not enabled because the in process node does not use the `config` property to configure the node, it expects `EXPERIMENTAL` to be in the root of the options object:

https://github.com/ipfs/js-ipfsd-ctl/blob/cf794cdb047c0ee1aaa8aff923c1fd0bd15e9051/src/ipfsd-in-proc.js#L36

dryajov:
Hmm, I believe that we can deprecate the use of options in `ipfsd-ctl` for `proc` types and always use the `config` key. That will make it consistent across all daemon types, which currently don't use options at all.

alanshaw:
Any progress on this? I've encountered it again...

jacobheun:
I created https://github.com/ipfs/js-ipfsd-ctl/pull/268 to resolve a separate issue I was having with `options.Bootstrap` not being overridable (which is painful when testing private networks). It does not do what @dryajov is proposing for the experimental args, but it does pass the config to the created IPFS node.

vasco-santos:
I would go with @jacobheun solution by now, and we can have @dryajov solution afterward. What do you think? @dryajov @alanshaw

daviddias:
@hugomrdias does this issue still apply?

hugomrdias:
not sure but i can guarantee in the next version this will not be a problem, already made tests for it

PR #395, opened 2019-10-11
feat: new ctl

hugomrdias:
Adds interface-core common tests setup to ipfsd-ctl so we can dedupe this code from js-ipfs (https://github.com/ipfs/js-ipfs/blob/master/test/utils/interface-common-factory.js#L71) and http-client (https://github.com/ipfs/js-ipfs-http-client/blob/master/test/utils/interface-common-factory.js).

Also adds jsdocs to be easier to transition to this new async setup. 

It works basically the same as the `createAsync` version with one less function call and a new method to create an unmanaged node to be used inside tests you want to stop manually.

PRs to use this in js-ipfs, http-client and interface-core will follow

This PR evolved from adding some helper functions for tests into a full rewrite (sorry about that 😅). While making the supporting PRs i found some problems and inconsistencies that needed to be fix to improve our own and contributors DX.

Problems: 
- Browsers tests skipped cause ctl didn't support proper connectivity to remote nodes
- We weren't able to tell ctl to use a specific commit of http-client, js-ipfs or cli
- Options/config between the 3 types of daemons weren't consistent
- Ctl didn't support remote "in process" daemon
- IPFS options were handled manually inside ctl, so any change in js-ipfs would require a PR in ctl to support the new options or change to an option

Related issues:
- https://github.com/ipfs/js-ipfsd-ctl/issues/208
- https://github.com/ipfs/js-ipfsd-ctl/issues/397
- https://github.com/ipfs/js-ipfsd-ctl/issues/374
- https://github.com/ipfs/js-ipfsd-ctl/issues/315
- https://github.com/ipfs/js-ipfsd-ctl/issues/207
- https://github.com/ipfs/js-ipfsd-ctl/issues/217
- and more 
 
Improvements:
- better errors
- DEBUG='ipfsd-ctl:*' everywhere
- factories for tests with good defaults
- options are properly merged everywhere
- safer child_process exit `stop()`
- faster stop()
- IPFS Options are now the same format as https://github.com/ipfs/js-ipfs/blob/master/README.md#ipfs-constructor
- Ctl can init, start and set config in one cmd (at least with js-ipfs)
- better docs and jsdocs
- we can now be sure which http-client, ipfs or go-ipfs is being used
- utils functions actually work in the browser now
- works in webworkers now
- simpler and faster overall
- disposable node actually clean themselves in the browser
- better tests
- ...
- support electron
- test in electron

New: 
- new method `createController` returns a spawned controller
- createFactory as a second parameter to override options per type

Changes: 
- `create` change to `createFactory`
- `createFactory` options changed   

Old
```md
- `options` - optional object with:
  - `remote` bool - use remote endpoint to spawn the nodes.
  - `port` number - remote endpoint port. Defaults to 43134.
  - `exec` - IPFS executable path. `ipfsd-ctl` will attempt to locate it by default. If you desire to spawn js-ipfs instances in the same process, pass the ref to the module instead (e.g `exec: require('ipfs')`)
  - `type` - the daemon type, see below the options
    - `go` - spawn go-ipfs daemon
    - `js` - spawn js-ipfs daemon
    - `proc` - spawn in-process js-ipfs instance. Needs to be called also with exec. Example: `DaemonFactory.create({type: 'proc', exec: require('ipfs') })`.
  - `IpfsClient` - A custom IPFS API constructor to use instead of the packaged one
```

**New**
```markdown
-   `remote` [boolean] Use remote endpoint to spawn the nodes. Defaults to `true` when not in node.
-   `test` [test=false] - Flag to activate custom config for tests.
-   `endpoint` [endpoint] - Endpoint URL to manage remote Controllers. (Defaults: 'http://localhost:43134').
-   `disposable` [Boolean] A new repo is created and initialized for each invocation, as well as cleaned up automatically once the process exits.
-   `type` [string] The daemon type, see below the options:-   go - spawn go-ipfs daemon
    -   js - spawn js-ipfs daemon
    -   proc - spawn in-process js-ipfs instance
-   `env` [Object] Additional environment variables, passed to executing shell. Only applies for Daemon controllers.
-   `args` [Array] Custom cli args.
-   `ipfsHttp` [Object] Setup IPFS HTTP client to be used by ctl.
    -   `ipfsHttp.ref` [Object] Reference to a IPFS HTTP Client object. (defaults to the local require(`ipfs-http-client`))
    -   `ipfsHttp.path` [string] Path to a IPFS HTTP Client to be required. (defaults to the local require.resolve('ipfs-http-client'))
-   `ipfsApi` [Object] Setup IPFS API to be used by ctl.
    -   `ipfsApi.ref` [Object] Reference to a IPFS API object. (defaults to the local require(`ipfs`))
    -   `ipfsApi.path` [string] Path to a IPFS API implementation to be required. (defaults to the local require.resolve('ipfs'))
-   `ipfsBin` [String] Path to a IPFS exectutable . (defaults to the local 'js-ipfs/src/bin/cli.js')
-   `ipfsOptions` [IpfsOptions] Options for the IPFS instance
```

- Previous default ipfs config is only applied when `test` options equals `true`
- `defaultAddrs` option was removed 
- Spawn options are the same as `createFactory`

Old
```
- `options` is an optional object the following properties:
  - `init` bool (default true) or Object - should the node be initialized
  - `initOptions` object - should be of the form `{bits: <size>}`, which sets the desired key size
  - `start` bool (default true) - should the node be started
  - `repoPath` string - the repository path to use for this node, ignored if node is disposable
  - `disposable` bool (default true) - a new repo is created and initialized for each invocation, as well as cleaned up automatically once the process exits
  - `defaultAddrs` bool (default false) - use the daemon default `Swarm` addrs
  - `args` - array of cmd line arguments to be passed to ipfs daemon
  - `config` - ipfs configuration options
```

**NEW**
Same as js-ipfs constructor https://github.com/ipfs/js-ipfs#ipfs-constructor
- ipfsd.killProcess removed not needed anymore
- ipfsd.getConfig removed call ipfsd.api.config.get instead
- ipfsd.setConfig removed, call ipfsd.api.config.set instead

TODO:
- [x] update readme
- [x] finish tests
- [x] update interface-core https://github.com/ipfs/interface-js-ipfs-core/pull/541
- [x] update http-client https://github.com/ipfs/js-ipfs-http-client/pull/1127
- [x] update js-ipfs https://github.com/ipfs/js-ipfs/pull/2529
- [ ] update interop
- [ ] update desktop @hacdias


src/factory-in-proc.js
```diff
@@ -4,84 +4,103 @@ const tmpDir = require('./utils/tmp-dir')
...
+  async version () {
+    const node = new InProc(merge({
+      ipfsOptions: {
+        start: true,
+        init: true,
+        config: defaultConfig,
+        repo: tmpDir(this.options.type)
+      }
+    }, this.options))
     const v = await node.version()
```
alanshaw: Shouldn't this new node be stopped? It's not used outside this function.

src/factory-daemon.js
```diff
@@ -48,48 +66,55 @@ class FactoryDaemon {
...
     // a way to get the version while the daemon is offline...
-    const d = new Daemon(options)
+    const d = new Daemon(merge({
+      ipfsOptions: {
+        start: true,
+        init: true,
+        config: defaultConfig,
+        repo: tmpDir(this.options.type)
+      }
+    }, this.options))
```
alanshaw: Should this be stopped after the version is retrieved?

src/index.js
```diff
@@ -53,21 +47,129 @@ const createServer = (options) => {
...
+      createOptions,
+      options
+    ))
+
+    // Spawn with merged options
+    const node = await ipfsFactory.spawn()
+
+    // Add `peerId`
+    const id = await node.api.id()
+    node.api.peerId = id
```
alanshaw: This is a lot less onerous when getting the ID is a one liner...it would be nice to not do this and would speed up the tests that don't actually use the peer ID.

src/ipfsd-daemon.js
```diff
@@ -35,41 +32,36 @@ function translateError (err) {
...
     this.apiAddr = null
     this.gatewayAddr = null
-    /** @member {IpfsClient} */
     this.api = null
   }
 
   setApi (addr) {
     this.apiAddr = multiaddr(addr)
-    this.api = (this.opts.IpfsClient || IpfsClient)(addr)
+    this.api = require(this.opts.ipfsHttp.path)(addr)
```
achingbrain: Why has this changed to require a path instead of invoking a constructor?

src/ipfsd-daemon.js
```diff
@@ -80,77 +72,56 @@ class Daemon {
...
-   * @param {Object} [initOptions={}]
-   * @param {number} [initOptions.bits=2048] - The bit size of the identiy key.
-   * @param {string} [initOptions.directory=IPFS_PATH] - The location of the repo.
-   * @param {string} [initOptions.pass] - The passphrase of the keychain.
+   * @param {Object} [initOptions={}] - @see https://github.com/ipfs/js-ipfs/blob/master/README.md#optionsinit
    * @returns {Promise}
    */
   async init (initOptions) {
-    if (this.initialized && initOptions) {
+    if (this.initialized && typeof initOptions === 'object') {
```
achingbrain: `initOptions` can now be a boolean or an object?  Why is that?

src/ipfsd-daemon.js
```diff
@@ -190,42 +165,29 @@ class Daemon {
...
-            resolve(this.api)
-          }
+        if (gwMatch && gwMatch.length > 0) {
+          this.setGateway(gwMatch[1])
+        }
+
+        if (output.match(/(?:daemon is running|Daemon is ready)/)) {
+          // we're good
+          this.started = true
+          resolve(this.api)
```
achingbrain: This will continue appending any logs to `output` after resolving the promise which causes a memory leak.

We should have two listeners for `stdout.on('data')` - one that calls `daemonLog.info(data.toString())` and the other that waits for the `Daemon is ready` type messages and unregisters itself when it before resolving the promise.

src/factory-in-proc.js
hugomrdias: yes this PR is still a WIP this is fixed already locally.
but im actually thinking about removing this method completely, not seeing much value in it plus we need to start the node to get the version, what do you think?

src/ipfsd-daemon.js
hugomrdias: we need both path and ref cause we can't send a ref over http to a remote daemon and the browser can't require from a path this is why the previous version was kinda random about using the correct http-client

src/ipfsd-daemon.js
hugomrdias: yes this version only supports options exactly in the same way js-ipfs does, ill rewrite the description of this PR explaining everything when i finish the tests

src/index.js
hugomrdias: theres a lot of tests that use api.peerId and lots of utils that do just this and api.id() is super simple impact should minimal.
to be fair i think we could do this in js-ipfs after init. 

@achingbrain didn't you find something related to this in the orbit-db codebase ?

alanshaw: Makes sense, lets leave it for now, we can always remove later.

src/ipfsd-daemon.js
achingbrain: Doesn't need to be fixed here though :)

achingbrain:
@hugomrdias are you going to polish off the last few nits in this?  It'd be good to get it in to unblock a bunch of other stuff.

alanshaw:
🙏 pretty please can we have this?

hacdias:
@alanshaw I sub you!

hugomrdias:
PR description updated explaining whats new and whats changed.

README.md
```diff
@@ -76,189 +76,164 @@ await server.stop()
...
-    - if `type` is 'js' a version string like `js-ipfs version: <version number>`
-     - if `type` is 'proc' an object with the following properties:
-        - version - the ipfs version
-        - repo - the repo version
-        - commit - the commit hash for this version
-
-### Remote endpoint - `const server = IPFSFactory.createServer([options])`
-
-`IPFSFactory.createServer` starts a IPFSFactory endpoint.
+### `create([options])`
```
achingbrain: This might be better named `createFactory([options])`?

README.md
```diff
@@ -76,189 +76,164 @@ await server.stop()
...
+Returns a **[TestsInterface](#TestsInterface)**
+#### TestsInterface
+`nodes` **Array** List of the created controlled nodes.   
+`node()` **Function** Creates a standalone node, this node will **NOT** be stopped by `teardown()`.   
+`setup(options)` **Function(**[FactoryOptions](#FactoryOptions)**)** Creates a controlled node.   
+`teardown()` **Function** Stops all controlled nodes created by `setup`   
 
-Get the address (multiaddr) of connected IPFS HTTP Gateway. Returns a multiaddr.
+### `createServer([options])`
+Create a Endpoint Server.
```
achingbrain:
```suggestion
Create an Endpoint Server.
```

src/endpoint/server.browser.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  /**
+   * Start the server
+   *
+   * @returns {Promise<Hapi.Server>}
+   */
+  start () {
+    console.warn('Server not implemented in the browser')
+    return Promise.resolve(this)
```
achingbrain: Should this (and the other methods in this class) throw instead?

README.md
hugomrdias: damn it was like that i just changed it back yesterday to reduce the diff for the end user.
do you think it's worth it ?

achingbrain: I think it's good to be explicit.  The diff is going to be hefty anyway, right?

hugomrdias: In the top level api nothing actually changed just the options are different

src/ipfsd-daemon.js
```diff
@@ -161,73 +145,80 @@ class Daemon {
...
+    args.push(...this.opts.args)
+
+    if (opts.pass && this.opts.type === 'js') {
+      args.push('--pass', '"' + opts.pass + '"')
+    }
+    if (opts.offline) {
+      args.push('--offline')
+    }
+    if (opts.preload && this.opts.type === 'js') {
+      args.push('--enable-preload', Boolean(opts.preload.enable))
```
alanshaw:
```suggestion
      args.push('--enable-preload', Boolean(opts.preload.enabled))
```

README.md
```diff
@@ -76,189 +76,164 @@ await server.stop()
...
+- `options` **[FactoryOptions](#FactoryOptions)** Factory options.
 
-**example:**
-```js
-const IPFSFactory = require('ipfsd-ctl')
+Returns a **[Factory](#factory)**
 
-const server = IPFSFactory.createServer({ port: 12345 })
+### `createNode([options])`
+Creates a node.
```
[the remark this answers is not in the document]
hugomrdias: `create` creates a factory and you get `.tmpDir()`, `.version()` and `.spawn()`

`createNode` return the node directly basically calls `spawn()` for you which is what you want 90% of the time.

i didn't made `createNode` default because there are some use cases that you really need `Factory.tmpDir()` to create a temp dir for the repo when using a remote node.

README.md
```diff
@@ -76,189 +76,164 @@ await server.stop()
...
+-   `env` **[Object]** Additional environment variables, passed to executing shell. Only applies for Daemon controllers.
+-   `args` **[Array]** Custom cli args.
+-   `ipfsHttp` **[Object]** Setup IPFS HTTP client to be used by ctl.
+    -   `ipfsHttp.ref` **[Object]** Reference to a IPFS HTTP Client object. (defaults to the local require(`ipfs-http-client`))
+    -   `ipfsHttp.path` **[string]** Path to a IPFS HTTP Client to be required. (defaults to the local require.resolve('ipfs-http-client'))
+-   `ipfsApi` **[Object]** Setup IPFS API to be used by ctl.
+    -   `ipfsApi.ref` **[Object]** Reference to a IPFS API object. (defaults to the local require(`ipfs`))
+    -   `ipfsApi.path` **[string]** Path to a IPFS API implementation to be required. (defaults to the local require.resolve('ipfs'))
+-   `ipfsBin` **[string]** Path to a IPFS exectutable . (defaults to the local 'js-ipfs/src/bin/cli.js')
+-   `ipfsOptions` **[IpfsOptions]** Options for the IPFS instance
```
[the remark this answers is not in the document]
hugomrdias: they work for all 3 types as far as possible exactly like the old version.

- `proc` its passed directly
- `daemon` it translates as much as possible to args (plus you can manually define and extra `args` option)
- `client` same as daemon cause its just a proxy

this change was to normalise and have just one set possible of options

README.md
```diff
@@ -76,189 +76,164 @@ await server.stop()
...
+-   `host` **[string]** Remote endpoint host. (Defaults to localhost)
+-   `port` **[number]** Remote endpoint port. (Defaults to 43134)
+-   `secure` **[string]** Remote endpoint uses http or https. (Defaults to false)
+-   `disposable` **[boolean]** A new repo is created and initialized for each invocation, as well as cleaned up automatically once the process exits.
+-   `type` **[string]** The daemon type, see below the options:-   go - spawn go-ipfs daemon
+    -   js - spawn js-ipfs daemon
+    -   proc - spawn in-process js-ipfs instance
+-   `env` **[Object]** Additional environment variables, passed to executing shell. Only applies for Daemon controllers.
+-   `args` **[Array]** Custom cli args.
+-   `ipfsHttp` **[Object]** Setup IPFS HTTP client to be used by ctl.
```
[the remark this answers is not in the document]
hugomrdias: nothing is mandatory, but if you really want to make sure you are using a specific version you need to specify both!
ill update the readme we that info !

README.md
```diff
@@ -76,189 +76,164 @@ await server.stop()
...
+Creates an interface to setup and teardown IPFS nodes for tests.
 
-Get the address (multiaddr) of connected IPFS API. Returns a multiaddr
+- `options` **[FactoryOptions](#FactoryOptions)** Factory options.
 
-#### `ipfsd.gatewayAddr` (getter)
+Returns a **[TestsInterface](#TestsInterface)**
+#### TestsInterface
+`nodes` **Array** List of the created controlled nodes.   
+`node()` **Function** Creates a standalone node, this node will **NOT** be stopped by `teardown()`.
```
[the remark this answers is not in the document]
hugomrdias: This needs better wording

`node()` returns the a ctl node 
`setup()` returns directly an ipfs core api 

Should both return the same thing ? 

i would rather add a new param than to use an option just to keep `FactoryOptions` clean. Cause this would be specific to the testInterface.


I also think using the word `node` here is wrong... We use daemon for process nodes and node for in process etc... but here it's actually a `Controller` for a IPFS node either process or in-proc.


Would you agree if i change the naming to:
`Controller` instead of `Node` ?

Then ipfsd-ctl would have:

`Factory` to wrap spawn, version and tmpDir
`Controller` to wrap the 3 different types of node
plus some methods that return an IPFS Core API.

README.md
```diff
@@ -76,189 +76,164 @@ await server.stop()
...
 
-Get the address (multiaddr) of connected IPFS API. Returns a multiaddr
+- `options` **[FactoryOptions](#FactoryOptions)** Factory options.
 
-#### `ipfsd.gatewayAddr` (getter)
+Returns a **[TestsInterface](#TestsInterface)**
+#### TestsInterface
+`nodes` **Array** List of the created controlled nodes.   
+`node()` **Function** Creates a standalone node, this node will **NOT** be stopped by `teardown()`.   
+`setup(options)` **Function(**[FactoryOptions](#FactoryOptions)**)** Creates a controlled node.
```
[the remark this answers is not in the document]
hugomrdias: yes

README.md
```diff
@@ -76,189 +76,164 @@ await server.stop()
...
+- `options` **[FactoryOptions](#FactoryOptions)** Factory options.
 
-console.log('endpoint is running')
+Returns a **[Node](#Node)**
 
-await server.stop()
-
-console.log('endpoint has stopped')
-```
+### `createNodeTests([options])`
```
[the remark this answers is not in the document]
hugomrdias: i can do that sure

hugomrdias:
Moved a few things around to integrate your feedback. Please review again.

README.md
```diff
@@ -76,189 +75,159 @@ await server.stop()
...
 
-**example:** See [Usage](#usage)
+- `options` **[ControllerOptions](#ControllerOptions)** Factory options.
 
-#### Get daemon version with `f.version() : Promise`
+Returns a **[Controller](#Controller)**
 
-Get the version without spawning a daemon
+### `createServer([options])`
+Create an Endpoint Server. This server is used by a client node to control a remote node. Example: Spawning a go-ipfs node from a browser.
```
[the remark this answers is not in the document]
hugomrdias: No, its mostly to be used by the remote `Controller` not for public use.

PR #396, opened 2019-10-14
fix: support more ipfs options

hugomrdias:
in-proc daemon didnt support all the js-ipfs options, this should be improved so we dont need to syncronize manually.

this is related to https://github.com/ipfs/interop/pull/86 circuit tests were failing because relay: true wasnt being passed to the constructor

### DIFF
--- a/src/ipfsd-in-proc.js
+++ b/src/ipfsd-in-proc.js
@@ -62,7 +62,11 @@ class InProc {
       EXPERIMENTAL: this.opts.EXPERIMENTAL,
       libp2p: this.opts.libp2p,
       config: this.opts.config,
-      silent: this.opts.silent
+      silent: this.opts.silent,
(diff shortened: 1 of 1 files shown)


achingbrain:
Do you think it might be worth putting the opts we pass to IPFS under a separate key in the opts passed to `InProc`?  No need to keep them in sync manually then.

hugomrdias:
> Do you think it might be worth putting the opts we pass to IPFS under a separate key in the opts passed to `InProc`? No need to keep them in sync manually then.

yes that is what i was thinking but that would be a breaking change, i wanna merge this and make a breaking version after.

achingbrain:
Can you open an issue for the breaking change so it doesn't get lost?

hugomrdias:
done https://github.com/ipfs/js-ipfsd-ctl/issues/397

PR #435, opened 2020-01-16
fix: remove swarm addrs in browser

alanshaw:
New libp2p is strict about providing addrs that you can listen on.

https://github.com/libp2p/js-libp2p/blob/83409deaa6773a550d38b77bd486faf8b8b97d29/src/transport-manager.js#L172-L176

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -4,7 +4,7 @@ const createServer = require('./src').createServer
 
 const server = createServer() // using defaults
 module.exports = {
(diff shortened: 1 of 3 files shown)


hugomrdias:
can you fix the bundlesize max value pls ?

hugomrdias:
ci is failing in the browser tests, looks like the tests need to be changed to support the changes.

Issue #438, opened 2020-01-20
Daemon is no longer force killed


alanshaw:
I'm seeing hangs for `factory.clean()` with Go IPFS nodes in the interop tests.

We used to have: https://github.com/ipfs/js-ipfsd-ctl/blob/7928c1eee092a9b96a59328294c84c9f065217a3/src/ipfsd-daemon.js#L263-L275

Would be good to get a warning when force killing so we know that it is happening.

PR #441, opened 2020-01-22
feat: force kill daemons after timeout

alanshaw:
This PR re-adds a feature that was removed in 1.0.0 whereby daemons are force killed with a SIGKILL when they do not stop within a timeout.

It adds 2 new options to the `Daemon` class - `forceKill` (boolean - default `true`) and `forceKillTimeout` (number - default 10000).

resolves https://github.com/ipfs/js-ipfsd-ctl/issues/438

### DIFF
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -243,6 +243,7 @@ class Daemon {
     let killed = false
     if (this.opts.forceKill !== false) {
       killTimeout = setTimeout(() => {
+        // eslint-disable-next-line no-console
         console.error(new Error(`Timeout stopping ${this.opts.type} node. Process ${this.subprocess.pid} will be force killed now.`))
         killed = true
         this.subprocess.kill('SIGKILL')
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -247,11 +247,6 @@ class Daemon {
         console.error(new Error(`Timeout stopping ${this.opts.type} node. Process ${this.subprocess.pid} will be force killed now.`))
         killed = true
 
-        if (this.opts.type === 'go') {
-          // will make go-IPFS print a stack trace to stderr
-          this.subprocess.kill('SIGQUIT')
-        }
-
         this.subprocess.kill('SIGKILL')
       }, this.opts.forceKillTimeout)
     }
@@ -261,7 +256,7 @@ class Daemon {
     } catch (err) {
       if (!killed) throw err // if was killed then ignore error
 
-      daemonLog.info('Daemon was killed')
+      daemonLog.info('Daemon was force killed')
     }
 
     clearTimeout(killTimeout)
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -246,14 +246,22 @@ class Daemon {
         // eslint-disable-next-line no-console
         console.error(new Error(`Timeout stopping ${this.opts.type} node. Process ${this.subprocess.pid} will be force killed now.`))
         killed = true
+
+        if (this.opts.type === 'go') {
+          // will make go-IPFS print a stack trace to stderr
+          this.subprocess.kill('SIGQUIT')
+        }
+
         this.subprocess.kill('SIGKILL')
-      }, this.opts.forceKillTimeout || 10000)
+      }, this.opts.forceKillTimeout)
     }
 
     try {
       await this.api.stop()
     } catch (err) {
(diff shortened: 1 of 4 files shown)


src/ipfsd-daemon.js
```diff
@@ -239,7 +239,24 @@ class Daemon {
       return this
     }
 
-    await this.api.stop()
+    let killTimeout
+    let killed = false
+    if (this.opts.forceKill !== false) {
```
hugomrdias: can you add default to true here https://github.com/ipfs/js-ipfsd-ctl/blob/3dae27670fdebe0edd6e1e26ddf5df4c274b62bf/src/factory.js#L20 and in the docs

src/ipfsd-daemon.js
```diff
@@ -239,7 +239,24 @@ class Daemon {
...
+      killTimeout = setTimeout(() => {
+        // eslint-disable-next-line no-console
+        console.error(new Error(`Timeout stopping ${this.opts.type} node. Process ${this.subprocess.pid} will be force killed now.`))
+        killed = true
+        this.subprocess.kill('SIGKILL')
+      }, this.opts.forceKillTimeout || 10000)
```
hugomrdias: can you add a default to 10000 here https://github.com/ipfs/js-ipfsd-ctl/blob/3dae27670fdebe0edd6e1e26ddf5df4c274b62bf/src/factory.js#L20 and in the docs

src/ipfsd-daemon.js
```diff
@@ -239,7 +239,24 @@ class Daemon {
...
+    }
+
+    try {
+      await this.api.stop()
+    } catch (err) {
+      if (!killed) throw err // if was killed then ignore error
```
hugomrdias: can you add a debug for the err when killed pls

src/ipfsd-daemon.js
```diff
@@ -239,7 +239,24 @@ class Daemon {
...
 
-    await this.api.stop()
+    let killTimeout
+    let killed = false
+    if (this.opts.forceKill !== false) {
+      killTimeout = setTimeout(() => {
+        // eslint-disable-next-line no-console
+        console.error(new Error(`Timeout stopping ${this.opts.type} node. Process ${this.subprocess.pid} will be force killed now.`))
+        killed = true
+        this.subprocess.kill('SIGKILL')
```
hugomrdias: can we use the this behaviour https://github.com/sindresorhus/execa#killsignal-options here ? send SIGTERM first and SIGKILL only if it doesnt die.

Stebalien:
Answering the question in https://github.com/ipfs/interop/pull/96#issuecomment-577617260, just send a SIGKILL. There's no real need to wait the 10s.

If someone wants to try to debug this:
* Wait for the test to hang.
* Send a SIGQUIT to go-ipfs.
* Look at the stacktrace (go-ipfs will send it to stderr on SIGQUIT).
  * Or file an issue with that stack trace.

hugomrdias:
@Stebalien thanks, so maybe we should reduce the timeout, wait 5s for `api.stop()` to finish after that SIGKILL immediately.

Stebalien:
SGTM!

achingbrain:
@hugomrdias I've pushed changes to this branch that address the PR comments (I think) - can it go in?

src/ipfsd-daemon.js
```diff
@@ -239,7 +239,32 @@ class Daemon {
...
+      }, this.opts.forceKillTimeout)
+    }
+
+    try {
+      await this.api.stop()
+    } catch (err) {
+      if (!killed) throw err // if was killed then ignore error
+
+      daemonLog.info('Daemon was killed')
```
hugomrdias:
```suggestion
      daemonLog.info('Daemon was force killed')
```

src/ipfsd-daemon.js
```diff
@@ -239,7 +239,32 @@ class Daemon {
...
+    if (this.opts.forceKill !== false) {
+      killTimeout = setTimeout(() => {
+        // eslint-disable-next-line no-console
+        console.error(new Error(`Timeout stopping ${this.opts.type} node. Process ${this.subprocess.pid} will be force killed now.`))
+        killed = true
+
+        if (this.opts.type === 'go') {
+          // will make go-IPFS print a stack trace to stderr
+          this.subprocess.kill('SIGQUIT')
+        }
```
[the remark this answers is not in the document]
achingbrain: Have removed it.

PR #449, opened 2020-01-29
fix: protect possible empty ipfsModule option

hugomrdias:
With this change only when `type === go` the option can be an empty object.

### DIFF
--- a/src/factory.js
+++ b/src/factory.js
@@ -25,6 +25,7 @@ const defaults = {
     path: require.resolve('ipfs-http-client'),
     ref: require('ipfs-http-client')
   },
+  ipfsModule: {},
   ipfsOptions: {},
   forceKill: true,
   forceKillTimeout: 5000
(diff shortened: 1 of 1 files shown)


src/factory.js
```diff
@@ -104,19 +105,19 @@ class Factory {
    * @returns {Promise<ControllerDaemon>}
    */
   async spawn (options = { }) {
+    const type = options.type || this.opts.type
     const opts = merge(
-      this.overrides[options.type || this.opts.type],
+      this.overrides[type],
+      // conditionally include ipfs based on which type of daemon we will spawn when none has been specifed
+      (type === 'js' || type === 'proc') ? {
```
achingbrain: What about when the user does:

```
factory.spawn({
  'type': 'js',
  ipfsModule: {
    path: require.resolve('some/other/path/to/ipfs'),
    ref: require('some/other/path/to/ipfs'),
  }
})
```

I think with this change the `ipfsModule` option the user passes in will get ignored

hugomrdias: no, user options is the last arg and overrides everything above

src/factory.js
```diff
@@ -104,19 +105,19 @@ class Factory {
...
+        ipfsModule: {
+          path: require.resolve('ipfs'),
+          ref: require('ipfs')
+        }
+      } : {},
       options
     )
 
-    // conditionally include ipfs based on which type of daemon we will spawn when none has been specifed
-    if ((opts.type === 'js' || opts.type === 'proc') && !opts.ipfsModule) {
```
achingbrain: I was thinking this needed to be something more like:

```javascript
if ((opts.type === 'js' || opts.type === 'proc') && !opts.ipfsModule) {	
  opts.ipfsModule = {}
}

if (opts.ipfsModule && !opts.ipfsModule.path) {
  opts.ipfsModule.path = require.resolve('ipfs')
}

if (opts.ipfsModule && !opts.ipfsModule.ref) {
  opts.ipfsModule.path = require(opts.ipfsModule.path)
}
```

hugomrdias: with the current changes we don't need to protect it like that, its already safe 

https://codesandbox.io/s/clever-ride-kr3r1

src/factory.js
achingbrain: Yeah, just thought of that :)

alanshaw: It still evaluates `require('ipfs')` which does not work when used in js-ipfs.

achingbrain: This is fixed in https://github.com/ipfs/js-ipfsd-ctl/pull/450

PR #450, opened 2020-01-29
fix: only require http api client if it has not been specified

achingbrain:
If we configure which http api module to use, don't try to require `ipfs-http-client`, because that then means the containing project needs to specify a dependency on `ipfs-http-client` when the containing project might actually be `ipfs-http-client`...

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -4,7 +4,7 @@ const createServer = require('./src').createServer
 
 const server = createServer() // using defaults
 module.exports = {
-  bundlesize: { maxSize: '930kB' },
+  bundlesize: { maxSize: '940kB' },
   karma: {
     files: [{
       pattern: 'test/fixtures/**/*',
--- a/.aegir.js
+++ b/.aegir.js
@@ -16,5 +16,12 @@ module.exports = {
   hooks: {
       pre: () => server.start(),
       post: () => server.stop()
+  },
+  webpack: {
+    externals: {
+      ipfs: 'ipfs',
+      'ipfs-http-client': 'ipfs-http-client',
+      'go-ipfs-dep': 'go-ifps-dep'
+    }
   }
 }
--- a/.aegir.js
+++ b/.aegir.js
@@ -17,11 +17,11 @@ module.exports = {
       pre: () => server.start(),
       post: () => server.stop()
   },
-  webpack: {
+  webpack: process.env.NODE_ENV === 'test' ? undefined : {
     externals: {
       ipfs: 'ipfs',
       'ipfs-http-client': 'ipfs-http-client',
-      'go-ipfs-dep': 'go-ifps-dep'
+      'go-ipfs-dep': 'go-ipfs-dep'
     }
   }
 }
--- a/.aegir.js
+++ b/.aegir.js
@@ -4,7 +4,7 @@ const createServer = require('./src').createServer
 
 const server = createServer() // using defaults
 module.exports = {
-  bundlesize: { maxSize: '940kB' },
+  bundlesize: { maxSize: '35kB' },
   karma: {
     files: [{
       pattern: 'test/fixtures/**/*',
--- a/package.json
+++ b/package.json
@@ -71,7 +71,10 @@
     "chai": "^4.2.0",
     "chai-as-promised": "^7.1.1",
     "dirty-chai": "^2.0.1",
+    "go-ipfs-dep": "^0.4.22",
     "husky": "^4.0.10",
+    "ipfs": "github:ipfs/js-ipfs#master",
+    "ipfs-http-client": "^42.0.0-pre.0",
     "lint-staged": "^10.0.2"
   },
   "peerDependencies": {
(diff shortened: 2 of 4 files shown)


hugomrdias:
hummm why did this jumped +9kb in size

achingbrain:
Transitive dependencies changed?

hugomrdias:
its 9kb gziped would need a lot random stuff from deps, can you run the aegir build analize please ? just to double check theses changes dont make us include http-client twice

achingbrain:
We shouldn't really be including `ipfs-http-client` at all since it's a peer dep.

I ran the analyser - it's quite hard to make out what's going on, they're definitely in there though only once that I can see.

Before:

image

After:

image

I've added them to webpack's externals list so they don't get added to the bundle.  Not sure if that's going to make the browser version explode at runtime or not.

achingbrain:
Lol:

image

hugomrdias:
its gonna break browser tests
i had a look and we have 3 node-forges plus a package.json in there so its **not** our fault.

please remove the last commit and lets merge/release this.

tip: use chrome to see the analyser, for some reason brave is broken.

.aegir.js
```diff
@@ -16,5 +16,12 @@ module.exports = {
   hooks: {
       pre: () => server.start(),
       post: () => server.stop()
+  },
+  webpack: {
+    externals: {
+      ipfs: 'ipfs',
+      'ipfs-http-client': 'ipfs-http-client',
+      'go-ipfs-dep': 'go-ifps-dep'
```
alanshaw:
```suggestion
      'go-ipfs-dep': 'go-ipfs-dep'
```

achingbrain: I'm so good at spelling this wrong

src/factory.js
```diff
@@ -115,17 +103,41 @@ class Factory {
   async spawn (options = { }) {
     const type = options.type || this.opts.type
     const opts = merge(
+      this.opts,
```
hugomrdias: overrides already have this.opts merged in

alanshaw:
Screenshot 2020-01-30 at 12 43 51

WAT! 🚀🐭

.aegir.js
```diff
@@ -4,7 +4,7 @@ const createServer = require('./src').createServer
 
 const server = createServer() // using defaults
 module.exports = {
-  bundlesize: { maxSize: '930kB' },
+  bundlesize: { maxSize: '940kB' },
```
hugomrdias:
```suggestion
  bundlesize: { maxSize: '35kB' },
```

package.json
```diff
@@ -71,7 +71,10 @@
     "chai": "^4.2.0",
     "chai-as-promised": "^7.1.1",
     "dirty-chai": "^2.0.1",
+    "go-ipfs-dep": "^0.4.22",
     "husky": "^4.0.10",
+    "ipfs": "github:ipfs/js-ipfs#master",
+    "ipfs-http-client": "^42.0.0-pre.0",
     "lint-staged": "^10.0.2"
   },
   "peerDependencies": {
```
hugomrdias: remove peer deps

package.json
```diff
@@ -71,7 +71,10 @@
     "chai": "^4.2.0",
     "chai-as-promised": "^7.1.1",
     "dirty-chai": "^2.0.1",
+    "go-ipfs-dep": "^0.4.22",
     "husky": "^4.0.10",
+    "ipfs": "github:ipfs/js-ipfs#master",
```
hugomrdias: remove github link

package.json
achingbrain: They are needed to run the tests with

src/factory.js
```diff
@@ -87,18 +74,39 @@ class Factory {
...
+    if (options.ipfsHttpModule) {
+      delete opts.ipfsHttpModule
+
+      if (options.ipfsHttpModule.path) {
+        opts.ipfsHttpModule = {
+          path: options.ipfsHttpModule.path
+          // n.b. no ref property - do not send code refs over http
         }
       }
+    }
```
hugomrdias: why do we need all this haven't we already done this below? we just need to remove refs like in the previous version of this block.

Plus its deleting and setting opts.ipfsModule where it should be opts.json.ipfsModule and same for opts.ipfsHttpModule, so if this code made some use case work it wasn't because of this logic.

achingbrain: It was resulting in `ipfsModule: {}` being sent over the http api which was then getting fleshed out to `ipfsModule: { path: require.resolve('ipfs'), ref: require('ipfs') }` which breaks under `js-ipfs`

hugomrdias: these if clauses are very hard to track and error prone because of all the combinations available in ctl

achingbrain: Fair point about the deletion of the wrong property though.

achingbrain: > these if clauses are very hard to track and error prone because of all the combinations available in ctl

You are quite right.  Here, look - I've removed them: https://github.com/ipfs/js-ipfsd-ctl/pull/454

PR #454, opened 2020-01-31
refactor: remove option normalisation

achingbrain:
There are just too many options, too many variables and too many different calling contexts so let's get the user to be explicit about how they intend to use the module rather than try to second guess them.

Also, allow the user to specify opts/overrides to the ipfsd server's use of `createFactory`.

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -1,8 +1,25 @@
 'use strict'
 
 const createServer = require('./src').createServer
+const { findBin } = require('./src/utils')
 
-const server = createServer() // using defaults
+const server = createServer(null, {
+  ipfsModule: {
+    path: require.resolve('ipfs'),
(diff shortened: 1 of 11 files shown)


achingbrain:
I can't say I'm coming across a lot of instances where I need one implementation of `ipfsModule` or `ipfsHttpModule` in one place and then another somewhere else so it doesn't make a lot of sense to allow them as options to `.spawn`.

If there's then no danger of sending refs over a http connection maybe we can get rid of the whole `path`/`ref` thing and just pass refs in to `createFactory`?

achingbrain:
@hugomrdias any changes needed here?

hugomrdias:
agree lets remove ref/path.

can we remove the second param from findBin also?

PR #457, opened 2020-02-09
refactor: remove path and ref from module args also findBin

achingbrain:
Since we're not passing modules to `.spawn()` there's no danger of passing module refs over http, so we can remove the `.path` and `.ref` arguments when specifying which `ipfs-http-client` and `ipfs` modules to use, instead setting them up when you configure the daemon factory, either in-proc or as a server in an `.aegir` file.

Also removes the `findBin` command magic in favour of being explicit about which binary you want the factory to use.

If you need to support env vars like `IPFS_GO_EXEC`, do that when you set up the factory, for example:

```javascript
const factory = createFactory({
  ipfsBin: process.env.IPFS_GO_EXEC || require('go-ipfs-dep').path()
})
```

Also fixes up the examples and upgrades the `ipfs` dep to run tests against the async/await refactor.

BREAKING CHANGES:

- `.path` and `.ref` args removed from `ipfsModule` and `ipfsHttpModule`
- `findBin` function removed

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -1,23 +1,16 @@
 'use strict'
 
 const createServer = require('./src').createServer
-const { findBin } = require('./src/utils')
 
 const server = createServer(null, {
-  ipfsModule: {
-    path: require.resolve('ipfs'),
-    ref: require('ipfs')
-  },
-  ipfsHttpModule: {
-    path: require.resolve('ipfs-http-client'),
-    ref: require('ipfs-http-client')
-  }
+  ipfsModule: require('ipfs'),
+  ipfsHttpModule: require('ipfs-http-client')
 }, {
   go: {
-    ipfsBin: findBin('go', true)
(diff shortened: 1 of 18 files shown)


achingbrain:
Done as a separate branch because `js-ipfs` master depends on the `remove-option-normalisation` branch currently.

hugomrdias:
given that js-ipfs depends in the target branch how do you want this merged ?

achingbrain:
I would merge the `remove-option-normalisation` PR into master but not delete the branch (since js-ipfs depends on it), then change the merge base for this PR to be master and merge it.

Next do a release from master and I'll PR js-ipfs to use the release version - once that's merged we can delete the `remove-option-normalisation` branch from this repo.

hugomrdias:
closing in favor https://github.com/ipfs/js-ipfsd-ctl/pull/458

PR #461, opened 2020-02-16
feat: auto fill factory options when it is unset

bluelovers:


hugomrdias:
Thank for your contribution, but we recently removed this type of logic because we found that is really hard and error prone to guess the best combination of require for every env. We feel that is best to be explicit and leave for the user to defined what they want to include.
So sadly im not gonna merge these changes.

PR #509, opened 2020-05-08
fix: remove onBlocked hook from delete database

achingbrain:
If we try to shut a node down while a query is still in flight it errors out.  The change here is to retry deleting the databse when we detect that it's still open, 5x attempts, each waiting 1-5s before trying again.  If it's still open after that, bail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "temp-write": "^4.0.0"
   },
   "devDependencies": {
-    "aegir": "^21.10.2",
+    "aegir": "^22.0.0",
     "benchmark": "^2.1.4",
     "go-ipfs-dep": "^0.5.0",
     "husky": "^4.2.5",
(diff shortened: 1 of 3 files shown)


achingbrain:
I think this is necessary because `close()` on IndexedDB is a lie:

> The close() method, when invoked, must run these steps:
>
> Run close a database connection with this connection.
>
> > The connection will not actually close until all outstanding transactions have completed. Subsequent calls to close() will have no effect.

https://w3c.github.io/IndexedDB/#dom-idbdatabase-close

hugomrdias:
its not enought to wait on success and remove the onblocked?

achingbrain:
I don't know, I assumed the `onblocked` handler there was for a reason?

hugomrdias:
looking into some libs this hook seems to be used for debug and is mostly optional. 
when i added the onblocked i think was more concerned with getting all the info possible than anything else. can you run the tests with it and see if it fails ?

achingbrain:
It seems happy enough; the tests pass so I've pulled out the retry and just wait for the onsuccess.

PR #520, opened 2020-07-10
fix: allow enabling dht in js-ipfs

achingbrain:
Turns the `EXPERIMENTAL.dht` option into the `--enable-dht` CLI flag.


achingbrain:
Actually, let's try to do without this.

PR #526, opened 2020-07-20
Always await daemon subprocess

### LINKED ISSUE #525: interface-ipfs-core with patched ipfsd-ctl 3.0.0 leaves a lot of PIPEWRAP behind
We recently had mac and windows builds start hanging on the "conformance" tests (just interface-ipfs-core for now). I've added why-is-node-running in https://github.com/rs-ipfs/rust-ipfs/pull/251 (on 60s unref'd timer once mocha finishes) and it seems that for 155 passing tests there are 90 PIPEWRAP instances (possibly) blocking node from exiting.

This is the log of the first run hanging run: https://github.com/rs-ipfs/rust-ipfs/pull/251/checks?check_run_id=889139927#step:17:288

* there are 90 PIPEWRAP which I think are for improperly shutdown nodes
* 1 DNSCHANNEL which might be the real reason for blocking

A PIPEWRAP looks like:

```
There are 91 handle(s) keeping the process running

# PIPEWRAP
D:\a\rust-ipfs\rust-ipfs\conformance\node_modules\execa\index.js:80                              - spawned = childProcess.spawn(parsed.file, parsed.args, parsed.options);
D:\a\rust-ipfs\rust-ipfs\conformance\node_modules\ipfsd-ctl\src\ipfsd-daemon.js:188              - this.subprocess = execa(this.exec, args, {
D:\a\rust-ipfs\rust-ipfs\conformance\node_modules\ipfsd-ctl\src\ipfsd-daemon.js:187              - const ready = new Promise((resolve, reject) => {
D:\a\rust-ipfs\rust-ipfs\conformance\node_modules\ipfsd-ctl\src\factory.js:142                   - await ctl.start()
D:\a\rust-ipfs\rust-ipfs\conformance\node_modules\interface-ipfs-core\src\miscellaneous\id.js:23 - ipfs = (await common.spawn()).api
```

The 2nd commit in the https://github.com/rs-ipfs/rust-ipfs/pull/251 ignores the test which creates the dns query which I guess is never reaped as of now.

Also I now notice that you've made a lot progress with this package and we are stuck with [patched 3.0.0](https://github.com/rs-ipfs/js-ipfsd-ctl/tree/add_rust_ipfs_http). Apologies up front if I cannot see the above being fixed in any of the changelog entries. We can try with a patched 5.0 to see if they get rid of those PIPEWRAPs.

I wonder if https://github.com/ipfs/js-ipfsd-ctl/pull/509 is related?

koivunej:
The idea is to always end up awaiting the subprocess even when the subprocess was killed. Not sure how correct the implementation is. Ok to supercede with another PR if you can see a better way to implement this. Pushing to run CI tests.

Fixes #525.

### DIFF
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -246,11 +246,21 @@ class Daemon {
       }, this.opts.forceKillTimeout)
     }
 
+    let stoppingErr
+
     try {
       await this.api.stop()
     } catch (err) {
+      // store for possible later use
+      stoppingErr = err
+    }
+
+    try {
+      // awaiting subprocesses should allow reaping any handles node vm has open for them
+      await this.subprocess
+    } catch (err) {
       if (!killed) {
-        throw err // if was killed then ignore error
+        throw stoppingErr || err // if was killed then ignore error
       }
 
       daemonLog.info('Daemon was force killed')
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -247,6 +247,7 @@ class Daemon {
     }
 
     let stoppingErr
+    let subprocessAwaitingErr
 
     try {
       await this.api.stop()
@@ -259,21 +260,30 @@ class Daemon {
       // awaiting subprocesses should allow reaping any handles node vm has open for them
       await this.subprocess
     } catch (err) {
-      if (!killed) {
-        throw stoppingErr || err // if was killed then ignore error
-      }
-
-      daemonLog.info('Daemon was force killed')
+      subprocessAwaitingErr = err
     }
 
-    clearTimeout(killTimeout)
+    // detach everything as we have done all of the teardown possible.
     this.subprocess.stderr.removeAllListeners()
     this.subprocess.stdout.removeAllListeners()
     this.started = false
 
+    clearTimeout(killTimeout)
+
+    if (killed) {
+      daemonLog.info('Daemon was force killed')
+    }
+
+    if (stoppingErr || subprocessAwaitingErr) {
+      if (!killed) {
+        throw stoppingErr || subprocessAwaitingErr // if was killed then ignore error
+      }
+    }
+
     if (this.disposable) {
       await this.cleanup()
     }
+
     return this
   }
 


koivunej:
I have only tested this locally on linux, where there are no issues, except that I dont have chrome.

koivunej:
Just realized the throw logic is still bad, of course the stop error should be rethrown if it happened but subprocess awaiting completed succesfully. Any ideas or examples to create a test case for this?

src/ipfsd-daemon.js
```diff
@@ -246,24 +246,44 @@ class Daemon {
...
+    if (stoppingErr || subprocessAwaitingErr) {
+      if (!killed) {
+        throw stoppingErr || subprocessAwaitingErr // if was killed then ignore error
+      }
+    }
+
     if (this.disposable) {
       await this.cleanup()
     }
+
```
vasco-santos, line 286:
```suggestion
```

src/ipfsd-daemon.js
```diff
@@ -246,24 +246,44 @@ class Daemon {
...
+      // store for possible later use
+      stoppingErr = err
+    }
 
-      daemonLog.info('Daemon was force killed')
+    try {
+      // awaiting subprocesses should allow reaping any handles node vm has open for them
+      await this.subprocess
+    } catch (err) {
+      subprocessAwaitingErr = err
```
hugomrdias, line 263: we already have a catch in the .start for the subprocess

koivunej: Took me a sec to understand, you probably meant this https://github.com/ipfs/js-ipfsd-ctl/blob/2851989eff489ef990e19ef8987d1c3b6c074aba/src/ipfsd-daemon.js#L214

Did some digging/asking around `Promise.prototype.catch` and if one tries this in node repl:

```
(async () => {
    try {
        let p = new Promise(resolve => setTimeout(resolve, 1000)).then(() => { 
            throw new Error("foo")
        });
        p.catch((e) => console.log("caught", e));
        await p
    } catch (e) {
        console.log("leaked", e);
    }
})()
```

The above example hits both console.log's which makes me think that the try-catch is required here.

Is the suggestion to just remove the try/catch from around this or what would it be? I still think the act of `await self.subprocess` here is important.

koivunej:
> I have only tested this locally on linux, where there are no issues, except that I dont have chrome.

To update my earlier comment, we have been running a fair amount of GHA builds since on linux/mac/windows without hangs -- [we](https://github.com/rs-ipfs/rust-ipfs/) [patch an earlier version of this](https://github.com/rs-ipfs/rust-ipfs/blob/master/conformance/patches/await_subprocess.patch) in [manually](https://github.com/rs-ipfs/rust-ipfs/blob/master/conformance/setup.sh#L19-L28) on top of our [3.0.0 fork](https://github.com/rs-ipfs/js-ipfsd-ctl/tree/add_rust_ipfs_http). There was of course the case of disabling some misc test case which may have triggered the actual hang, but however this change was necessary to "guess better" the hanging test. Did not do any analysis on *why* it was hanging though (I can only handle so much WIP :)).

src/ipfsd-daemon.js
achingbrain: In your example above if you reassign the value of p then the error is only printed once:

```js
(async () => {
    try {
        let p = new Promise(resolve => setTimeout(resolve, 1000)).then(() => { 
            throw new Error("foo")
        });
        p = p.catch((e) => console.log("caught", e));   // <--- reassign
        await p
    } catch (e) {
        console.log("leaked", e);
    }
})()

Promise { <pending> }
> caught Error: foo
    at repl:4:19
    at async repl:7:9
```

..which is why the second try-catch is necessary.

Also this code:

```js
this.subprocess.catch(err => reject(translateError(err))) 
```

will just reject the `ready` promise created & awaited on in `.start` which by this point has already been resolved so the error is effectively swallowed.

koivunej: Yeah, did consider this originally but I understood from MDN that this would make the reassigned `p` (return value from `Promise.prototype.catch`) resolve only if the original promise was rejected, however this didn't seem to be the case when testing this.

So to recap:

 1. Reassign the this.subprocess to be the promise returned by the `this.subprocess.catch(err => reject(...)))` on line 214
 2. Remove second try-catch from `stop`

I'll try to do these later today.

achingbrain:
Is this PR still necessary?  The code behind the daemon processes has been refactored, could you check with the latest version and see if the problem this PR solves is still apparent?

koivunej:
Apologies for forgetting about this PR.

I still see a lot of hangs in 7.1.0 without other patches than the `IPFS_RUST_EXEC`, actually more than before. It seems that all of the executions hang except for the miscellanius.stop test which calls shutdown manually. I wnoder if the "add rust-ipfs support" was incomplete in the beginning, or the latest disposable changes have changed this...? Trying to track it down, perhaps there should be some `{ disposable: true }` somewhere..

On 7.1.0 if I capture all of the logs with `DEBUG=*` and run all tests, I can only see `POST /api/v0/shutdown` happening from the miscellanious.stop test case. I wonder if there is some other mechanism that we should be stopping instead of `POST /api/v0/shutdown`?

Repeating the same for our old, very-much patched 3.0.0 version we've been using for a long time does consistently call `POST /api/v0/shutdown` when doing cleanup, and because of patching, awaits for subprocess and so never hangs.

koivunej:
Sprinkling some debugging all over src/ipfsd-daemon.js reveals me that the defaults were correct, this line gets executed and everything is well (except the process was not killed, no exception are thrown, `this.subprocess.on("exit", () => ...)` is executed): https://github.com/ipfs/js-ipfsd-ctl/blob/8fa51ce495934a751dc2b91fd9e86af56b669f96/src/ipfsd-daemon.js#L231

Ah, seems that the issue was with our wrapper shell script used to capture logs :) it was successfully killed but doing so left the child process spawned by the wrapper alive. Oh well, curious self inflicted issue there. I was wondering why it seemed like those were orphaned processes.

I'll update/close this after I can confirm no more hangs without the unnecessary wrapper.

koivunej:
Interestingly enough I couldnt get a hang locally after removing the wrapper script but a lot of PIPEWRAP still on github actions ubuntu-latest: https://github.com/rs-ipfs/rust-ipfs/pull/425/checks?check_run_id=1348215782#step:19:5699

32 handles doesn't sound familiar but it's possible this has now shifted some of the errors somewhere else. I'll try patching in some subprocess await and logging somewhere and see if I can continue debugging this on github actions.

koivunej:
I've done quite a bit of digging related to this and feel stuck. Luckily the issue is repeatable on my local laptop and github-actions `ubuntu-latest`. The issue seems that `p-wait-for` conditions do not seem to be rescheduled. Longer explanation follows:

Possibly unrelated or misdiagnosed: This seems like a re-entrancy hazard but didn't solve any random lockups (can be avoided by creating many factories for the tests). For "hazard" I was thinking that there might be two calls to `common.cleanup()` which end up cleaning the same instances, but that probably shouldn't happen. https://github.com/ipfs/js-ipfsd-ctl/blob/8fa51ce495934a751dc2b91fd9e86af56b669f96/src/factory.js#L156-L160

By adding even more logging it seems that the code gets stuck on with nodejs 14.x here: https://github.com/ipfs/js-ipfsd-ctl/blob/8fa51ce495934a751dc2b91fd9e86af56b669f96/src/ipfsd-daemon.js#L253-L255

<details>
<summary>
Log summary of one of the more popular cases to hang:
<ul>
<li>6 x `begin removing`</li>
<li>(5 + 1 later) x `awaiting for disposing`</li>
<li>6 x `subprocess cleanup complete` (logged at the end of `ipfsd-daemon.js:cleanup`)</li>
<li>10 x `awaiting for disposing`</li>
<li>only 5 x `stop completed`, one remains but does not seem to be polled</li>
</ul>
</summary>

```
  .swarm.peers
    ✓ should respect timeout option when listing swarm peers
    ✓ should list peers this node is connected to
    ✓ should list peers this node is connected to with verbose option
    ✓ should list peers only once (1393ms)
    ✓ should list peers only once even if they have multiple addresses (1290ms)
  ipfsd-ctl:factory cleaning up 6 controllers +3s
  ipfsd-ctl:daemon begin removing /tmp/rust_ipfs_OAi3CY-ebfm88FnPuvW5n +3s
  ipfsd-ctl:daemon awaiting for disposing +0ms
  ipfsd-ctl:daemon begin removing /tmp/rust_ipfs_v4EfcbNikiZzsblZQK4eO +1ms
  ipfsd-ctl:daemon awaiting for disposing +1ms
  ipfsd-ctl:daemon begin removing /tmp/rust_ipfs_r1lmLGGxlRhcVPtAo8gRO +0ms
  ipfsd-ctl:daemon awaiting for disposing +1ms
  ipfsd-ctl:daemon begin removing /tmp/rust_ipfs_h6v4MowkhUsUMrwNMxxt5 +0ms
  ipfsd-ctl:daemon awaiting for disposing +1ms
  ipfsd-ctl:daemon begin removing /tmp/rust_ipfs_noqaWwhoVzdrlWvFknuPV +0ms
  ipfsd-ctl:daemon awaiting for disposing +1ms
  ipfsd-ctl:daemon begin removing /tmp/rust_ipfs_RwiKoeKQaxQBYSqpZac0d +0ms
  ipfsd-ctl:daemon subprocess cleanup complete +7ms
  ipfsd-ctl:daemon subprocess cleanup complete +0ms
  ipfsd-ctl:daemon subprocess cleanup complete +0ms
  ipfsd-ctl:daemon subprocess cleanup complete +0ms
  ipfsd-ctl:daemon subprocess cleanup complete +1ms
  ipfsd-ctl:daemon subprocess cleanup complete +0ms
  ipfsd-ctl:daemon awaiting for disposing +8ms
  ipfsd-ctl:daemon stop completed +1ms
  ipfsd-ctl:daemon awaiting for disposing +0ms
  ipfsd-ctl:daemon stop completed +1ms
  ipfsd-ctl:daemon awaiting for disposing +0ms
  ipfsd-ctl:daemon stop completed +0ms
  ipfsd-ctl:daemon awaiting for disposing +1ms
  ipfsd-ctl:daemon stop completed +1ms
  ipfsd-ctl:daemon awaiting for disposing +0ms
  ipfsd-ctl:daemon stop completed +1ms
    1) "after all" hook for "should list peers only once even if they have multiple addresses"
```

`awaiting for disposing` is the condition fn at https://github.com/ipfs/js-ipfsd-ctl/blob/8fa51ce495934a751dc2b91fd9e86af56b669f96/src/ipfsd-daemon.js#L245
</details>

It doesnt seem like the polling would consistently happen on every interval, but I cannot find any obvious issues in `p-wait-for`. It is implemented exactly as I'd expect it to be.

I've pestered some folks internally at eq about this, in additions to github actions (windows-latest, macosx-latest) it doesn't reproduce on:
 - ubuntu 20.04 with nodejs 12, intel
 - arch with nodejs 15, ryzen

I get a hang nearly every time on ubuntu 19.10 based with nodejs {12,13,14} on my ryzen 5 laptop, 5.3.0 kernel. Running with NODE_DEBUG=timer doesn't really reveal anything useful. Reproduces on cpufreq governor ondemand and performance.

This feels like a "Select is broken" kind of situation but I haven't been able to pinpoint this better than ordering of calls from `p-wait-for` to the `conditional`. On successful runs it'd seem that the callbacks are executed in favourable order as in, polling is done right after execution but on hanging runs like above, polling seemingly stops.

Note: there may very well be actual bugs and so on, but the hang similar to above folded log looks like: `1) "after all" hook in ".pubsub.subscribe"`. The timeout is large enough to check that there are in fact zero rust-ipfs processes running at that time I consider it hung.

<details>
<summary>Reproduction instructions, code review</summary>

1. install rustup using https://rustup.rs/ (it self-updates if you have installed it previously)
2. do `rustup update stable`
3. git clone https://github.com/koivunej/rust-ipfs (note different from previous repro attempts)
4. in rust-ipfs working copy `git checkout -b origin/upgrade_js_ipfsd_ctl`
5. in rust-ipfs working copy `cargo build -p ipfs-http` (this will take a while)
6. in other terminal, in rust-ipfs working copy go under `conformance/`, start `./setup.sh` (you'll need to run it from `conformance/`)
7. wait for both `setup.sh` and `cargo -p ipfs-http` to complete

to get the few test cases being stuck, in `conformance/` run `DEBUG=ipfsd-ctl:daemon,ipfsd-ctl:factory IPFS_RUST_EXEC="$(pwd)/http" npm test` (the `conformance/http` is symlink created by `setup.sh`)

The code being used is js-ipfsd-ctl from https://github.com/koivunej/js-ipfsd-ctl/tree/add_rust_ipfs_http_on_7_wip
</details>

lidel:
What is the status of this folks?
Ok to close, or salvageable?

koivunej:
Sorry for leaving this open. It seems that there has been new developments, at least a single conflicts worth.

I was never really able to understand why this failed on my for me locally, on multiple nodejs versions, but it wasn't reproducable. I'll close this for now and reopen if I ever get time to investigate further.

PR #542, opened 2020-08-29
feat: add --migrate

bluelovers:


achingbrain (changes requested):
Can you add a test for this please.

As the configuration matrix gets more complicated we'll need some way of verifying functionality like this.

bluelovers:
how can i add test when current not exists any test for options

it is not a shared function

achingbrain:
I would refactor the creation of the `args` array into a method on the `Daemon` class called `_buildArgs(opts)`.

Invoke that method in the test an assert that the correct arg is added for the given option.

bluelovers:
so can u merge this, any maybe one day this repo will splitting code for make it reuse/test able

achingbrain:
No, I think new code needs tests, otherwise we cannot guarantee there will be no regressions in future.  Please add a test, I've outlined how to do it.

bluelovers:
@achingbrain ok, refactor done

achingbrain:
CI is failing on this PR, could you please take a look?

jacobheun:
@bluelovers can you take a look at the CI failures?

achingbrain:
Will be landed in #542 as this PR has been abandoned.

bluelovers:
@achingbrain do not ask change code out of my pr want to do at next time, thx

PR #549, opened 2020-09-26
feat: go-ipfs 0.7.0 compatibility

achingbrain:
Removes `bits` option when initting go-ipfs repos as the move to ed25519 keys makes it throw when `bits` is passed.


aschmahmann:
Note that you probably are already aware of: you can still use bits when initting with RSA keys instead of the default Ed25519 keys.

PR #554, opened 2020-10-22
fix: kill disposable nodes on stop and simplify started status

achingbrain:
Fixes two problems:

1. If we use `ipfs-daemon` to connect to a running node, there's no `this.subprocess` only `this.api` so don't reference that property in the `this.stop()` method.
2. There's a delay between stopping the process and the process actually stopping - in CI we don't wait for this and start new tests which start new processes which can overwhelm small weak build machines so wait for the subprocess to stop before returning from `this.stop()`

Changes behaviour:

1. Let the exiting of `this.subprocess` set `this.started = false` instead of when we stop the process
2. If a node is marked `disposable`, we don't really care about it exiting cleanly as we're going to delete it's repo shortly afterwards so just `SIGKILL` it immediately.

### DIFF
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -276,20 +276,12 @@ class Daemon {
         this.subprocess.cancel()
       }
 
-      try {
-        await this.subprocess
-      } catch (err) {
-        if (!this.subprocess.killed) {
-          throw err
-        }
-      } finally {
-        clearTimeout(killTimeout)
-      }
-
       // wait for the subprocess to exit and declare ourselves stopped
       await waitFor(() => !this.started, {
         timeout
       })
+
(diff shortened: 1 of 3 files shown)


src/ipfsd-daemon.js
```diff
@@ -228,43 +247,55 @@ class Daemon {
   /**
    * Stop the daemon.
    *
+   * @param {object} options
+   * @param {number} options.timeout - How long to wait for the daemon to stop
```
jacobheun, line 251: nit: specify as optional and include default
```suggestion
   * @param {object} [options]
   * @param {number} [options.timeout=60000] - How long to wait for the daemon to stop
```

jacobheun (approved):
Just a nit, otherwise this LGTM

PR #555, opened 2020-10-23
docs: update electron example

achingbrain:
Adds configuration to server creation and updates deps


jacobheun (approved):
LGTM, verified this built and ran locally.

PR #616, opened 2021-03-21
chore: use `.create` function for http client and ipfs client

achingbrain:
ipfs/js-ipfs#3550 switches away from default exports for `ipfs-http-client`
and `ipfs-client` to enable an easier transition to es6 modules.

All `ipfs` modules now export a `.create` factory function which returns
instances of the client module.

BREAKING CHANGE: expects `ipfs-http-client` and `ipfs-client` to export a `.create` function

### DIFF
--- a/src/ipfsd-client.js
+++ b/src/ipfsd-client.js
@@ -76,12 +76,12 @@ class Client {
    */
   _createApi () {
     if (this.opts.ipfsClientModule && this.grpcAddr && this.apiAddr) {
-      this.api = this.opts.ipfsClientModule({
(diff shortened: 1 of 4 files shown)


hugomrdias (approved):

achingbrain:
This PR needs https://github.com/ipfs/js-ipfs/pull/3550 for the build to pass due to the `ipfs-http-client` exports change, which needs this PR for it's build to pass.  I'm going to merge this an release as a major, then we can circle back here once https://github.com/ipfs/js-ipfs/pull/3550 is shipped.

PR #623, opened 2021-04-06
chore: remove peer deps

achingbrain:
We added ipfs, the http client etc as peer deps to signal to the user
that they should add the modules they depend on as deps of their project.

Starting with npm7 all peer deps get installed automatically which
defeats the purpose of our use of peer deps, so let's remove them.

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -1,13 +1,24 @@
 'use strict'
 
+const path = require('path')
 const getPort = require('aegir/utils/get-port')
 const createServer = require('./src').createServer
+
+/** @type {import('aegir').Options["build"]["config"]} */
+const esbuild = {
+  inject: [path.join(__dirname, 'scripts/node-globals.js')],
+}
 
 module.exports = {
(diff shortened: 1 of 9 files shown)


achingbrain:
Also removes the pre-commit hooks because we don't use them any more, and the thing just caused me to lose a commit message as it errored out trying to run a command.

hugomrdias (approved):

achingbrain:
Getting some corrupt tarballs while installing deps in CI:

```
error registry.npmjs.org/@leichtgewicht/readable-stream/-/readable-stream-3.6.0.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "ENOENT: no such file or directory, chmod '/home/runner/.cache/yarn/v6/.tmp/6fb5c27151ebc79d92adfda8e6b6674c/LICENSE'"
```

achingbrain:
Not massively helpful: https://github.com/yarnpkg/yarn/issues/6312

Issue #468, opened 2020-02-22
sometime ipfs api file didn't delete when server stoped


bluelovers:
when `IPFD_PATH/api` file exists, Controller will never start and throw error

hope can fix this

the code for how i avoid this, but hope can have build-in way fix this
https://github.com/bluelovers/ws-ipfs/blob/master/packages/use-ipfs/lib/ctl.ts

hugomrdias:
why are starting a new ipfs-http-client ? i really don't understand the situation , can explain further?

bluelovers:
it for test ipds is running or not

hugomrdias:
i dont understand what you mean

bluelovers:
hmm... i think u just need understand 

`IPFD_PATH/api` sometime didn't delete when ipfs is stop
it make `js-ipfsd-ctl` can't start ipfs

don't care how i try avoid this happen

my point only

1. check `IPFD_PATH/api` is exists or not
2. if it exists, check ipfs is running or not
3. if ipfs is not running, delete `IPFD_PATH/api`
4. start ipfs

hugomrdias:
I cannot reproduce this, just think about it the tests wouldn't work if this happened, ipfs tests use ctl would not work also, ipfs-http-client tests would not work.
Either you can produce a **minimal** repo where this happens or I'm gonna close this issue.

bluelovers:
it is random happen

PR #653, opened 2021-07-30
fix(daemon): avoid ipfs api file is exists when server is not running

bluelovers:
https://github.com/ipfs/js-ipfsd-ctl/issues/468


lidel:
@bluelovers are you able to add a test case that demonstrates what is fixed by this change?

PR #654, opened 2021-07-30
feat(api): support global timeout

bluelovers:


lidel:
@bluelovers mind adding tests that confirm timeout works?

bluelovers:
> @bluelovers mind adding tests that confirm timeout works?

i don't know how to do test

PR #722, opened 2022-03-14
fix: pass ipfs config show errors

### LINKED ISSUE #721: Need to trap JSON.parse and emit more informative error on parse failure
https://github.com/ipfs/ipfs-desktop/issues/2004 and https://github.com/ipfs/ipfs-desktop/issues/2000 show that a JSON parse error happened, which points to https://github.com/ipfs/js-ipfsd-ctl/blob/70dff97afc5a194afc106b817b98575f8b122723/src/ipfsd-daemon.js#L366 

we should wrap the json.parse in a try/catch and provide a more informative error message:

1. What is the value of `stdout`
2. What is the value of `key`

lidel:
Closes #721

### DIFF
--- a/src/ipfsd-daemon.js
+++ b/src/ipfsd-daemon.js
@@ -353,7 +353,8 @@ class Daemon {
    */
   async _getConfig (key = 'show') {
     const {
-      stdout
+      stdout,
+      stderr
     } = await execa(
       this.exec,
       ['config', key],
@@ -363,7 +364,12 @@ class Daemon {
       .catch(translateError)
 
     if (key === 'show') {
-      return JSON.parse(stdout)
+      try {
+        return JSON.parse(stdout)
+      } catch (err) {
+        err.message = `${err.message}: ${stderr || stdout}`
+        throw err
+      }
     }
 
     return stdout.trim()


lidel:
@SgtPooki  mind pushing this over the finish line? I think it needs some typescript definitions

BigLep:
@SgtPooki : please also add auto-release

src/ipfsd-daemon.js
```diff
@@ -363,7 +364,12 @@ class Daemon {
       .catch(translateError)
 
     if (key === 'show') {
-      return JSON.parse(stdout)
+      try {
+        return JSON.parse(stdout)
+      } catch (err) {
```
achingbrain, line 369:
```suggestion
      } catch (/** @type {any} */ err) {
```

achingbrain (changes requested):

src/ipfsd-daemon.js
SgtPooki: Why type any and not unknown

SgtPooki: Or Error&unknown

SgtPooki (changes requested):
Is this PR still needed? Idk how this slipped by me but I can take this tomorrow

PR #761, opened 2022-08-09
fix: make sure to behave as node.js in electron main process

### LINKED ISSUE #760: doesn't properly work in Electron
As we know in electron there two processes: renderer (browser) and main (node.js). When I use library in main process I expect it to behave the same way as I used in node.js. However it doesn't work this way and it requires "server" (I'm talking about createServer) to be created. Looks like it treats a main process as browser and tries to work like in a browser?

pleerock:
Fixes #760; electron main process can be treated as node.js, so we don't need to use remote server.


achingbrain:
Thanks!

PR #764, opened 2022-08-19
feat: add js-kubo-rpc-client

SgtPooki:
see https://github.com/ipfs/kubo/issues/9125 and https://github.com/ipfs/js-kubo-rpc-client/issues/3

fixes https://github.com/ipfs/js-kubo-rpc-client/issues/3


.aegir.js
```diff
@@ -20,10 +21,19 @@ export default {
...
+       */
+      let controllerOptions = {
+        ipfsModule,
+      }
+
+      if (process.env.USE_KUBO_JS) {
+        controllerOptions.kuboRpcModule = kuboRpcModule
+      } else {
+        controllerOptions.ipfsHttpModule = ipfsHttpModule
+      }
```
lidel, line 35: @SgtPooki  Why do we need `USE_KUBO_JS`? We don't have that for embedded JS-IPFS (`ipfsModule`).

SgtPooki: this is just inside .aegir.js, so it shouldn't be required by consumers. This is simply to toggle which module we're using when running tests. This PR includes changes that will use whichever option is passed in.

SgtPooki: See the diff, but here's links to the changes:

https://github.com/ipfs/js-ipfsd-ctl/blob/7fa9e7e8272ae4cab4e068a88ed9be8a815186fb/src/ipfsd-client.js#L97-L105
https://github.com/ipfs/js-ipfsd-ctl/blob/7fa9e7e8272ae4cab4e068a88ed9be8a815186fb/src/ipfsd-daemon.js#L103-L110
https://github.com/ipfs/js-ipfsd-ctl/blob/7fa9e7e8272ae4cab4e068a88ed9be8a815186fb/src/ipfsd-in-proc.js#L71-L80

SgtPooki: Eventually, we would remove USE_KUBO_JS (maybe we do so in this PR? lmk) and only run tests against kuboRpcModule, but I thought during the transition phase that we should be testing against both. open to either option

SgtPooki:
@lidel I removed the USE_KUBO_JS env var and just passed as overrides to ipfsd-ctl now. should be good to go on this PR

achingbrain (approved):

achingbrain (changes requested):
I think it's worth adding a test that uses the kubo rpc client

lidel (dismissed):
iiuc  `test/*.spec.js` still use `js-ipfs-http-client` for talking to Kubo? (e.g., controller one)
I think you need to either refactor them to use js-kubo-rpc-client   instead, or duplicate, so we  test both.

SgtPooki:
Good catch, i'm not sure how I missed that. Updated and added logging to ensure that we're using the module we think we're using.

SgtPooki:
Log of running `DEBUG='ipfsd-ctl:*rpcModule*' npm run test &> test.log.txt` [test.log.txt](https://github.com/ipfs/js-ipfsd-ctl/files/9619784/test.log.txt)

SgtPooki:
> iiuc `test/*.spec.js` still use `js-ipfs-http-client` for talking to Kubo? (e.g., controller one) I think you need to either refactor them to use js-kubo-rpc-client instead, or duplicate, so we test both.

There are already tests using kubo via ipfsd-ctl's `type: 'go'` so I just updated controller.spec.js so that it's using kuboRpcModule now. That should be sufficient.. I don't think adding a new test will gain us anything right now because they're essentially the same packages.

src/ipfsd-daemon.js
```diff
@@ -19,6 +19,7 @@ const daemonLog = {
   info: logger('ipfsd-ctl:daemon:stdout'),
   err: logger('ipfsd-ctl:daemon:stderr')
 }
+const rpcModuleLogger = logger('ipfsd-ctl:daemon:rpcModule:stdout')
```
achingbrain, line 22: The `stdout`/`stderr` suffix is because we're logging the output of an external process, that's not the case for this logger so it should be shortened:

```suggestion
const rpcModuleLogger = logger('ipfsd-ctl:daemon')
```

src/ipfsd-client.js
```diff
@@ -9,6 +9,7 @@ const daemonLog = {
   info: logger('ipfsd-ctl:client:stdout'),
   err: logger('ipfsd-ctl:client:stderr')
 }
+const rpcModuleLogger = logger('ipfsd-ctl:client:rpcModule:stdout')
```
achingbrain, line 12: The `stdout`/`stderr` suffix is because we're (in theory) logging the output of an external process, that's not the case for this logger so it should be shortened:

```suggestion
const rpcModuleLogger = logger('ipfsd-ctl:client')
```

package.json
```diff
@@ -159,6 +159,7 @@
     "execa": "^6.1.0",
     "ipfs-utils": "^9.0.1",
     "joi": "^17.2.1",
+    "kubo-rpc-client": "^1.0.0",
```
achingbrain, line 162: This should be a dev dep the same as `ipfs`, `ipfs-client` and `ipfs-http-client`, otherwise you won't be able to use this module to test with semver-incompatible versions of `kubo-rpc-client`.

README.md
```diff
@@ -62,17 +62,15 @@ $ npm i ipfsd-ctl
 
 Version 1.0.0 changed a bit the api and the options methods take so please read the documentation below.
 
-Please ensure your project also has dependencies on `ipfs`, `ipfs-http-client` and `go-ipfs`.
+Please ensure your project also has dependencies on `ipfs`, `ipfs-http-client`, `kubo-rpc-module`, and `go-ipfs`.
```
achingbrain, line 65: Incorrect module name:

```suggestion
Please ensure your project also has dependencies on `ipfs`, `ipfs-http-client`, `kubo-rpc-client`, and `go-ipfs`.
```

achingbrain (changes requested):
LGTM, just a few small tweaks required

achingbrain:
Have you tried building js-ipfs (or anything with types) with this change?  Kind of curious if the [`.api`](https://github.com/ipfs/js-ipfsd-ctl/blob/master/src/types.ts#L28) property is going to need some more attention for when/if `kubo-rpc-client` breaks compatibility.

SgtPooki:
> Have you tried building js-ipfs (or anything with types) with this change? Kind of curious if the controller [.api](https://github.com/ipfs/js-ipfsd-ctl/blob/master/src/types.ts#L28) property is going to need some more attention for when/if kubo-rpc-client breaks compatibility.

I've built the types in https://github.com/ipfs/interop/pull/499 but not js-ipfs.

achingbrain:
I've done the typescript conversion work in #777 (like #776 but without the changes from this PR).

I don't want to merge it off the bat as it'll stomp all over these changes - could you try integrating `kubo-rpc-client` in a PR targeting the branch from #777 - it'll likely give us a bit more confidence over the types as the jsdoc stuff is pretty limited.

package.json
```diff
@@ -152,13 +152,15 @@
     "wherearewe": "^2.0.1"
   },
   "devDependencies": {
+    "@libp2p/interfaces": "^3.0.3",
```
SgtPooki, line 155: without this, `ipfs-core-types` was failing on a lack of types from `@libp2p/interfaces/events` from the line `import type { EventHandler } from '@libp2p/interfaces/events'`

SgtPooki:
tests seem to be stuck on this PR for some reason.. running them at https://github.com/SgtPooki/js-ipfsd-ctl/actions/runs/3178121664

SgtPooki:
@achingbrain CI is passing

BigLep:
@SgtPooki : that's great.

@achingbrain : any concerns with merging?  I want get @SgtPooki unblocked.  If it's not ok to merge, can you lay out what the conditions for merging will be?

package.json
achingbrain, line 155: This dep is added but doesn't appear to be used anywhere?

```suggestion
```

SgtPooki: When removing this, ipfs-core-types throws an error

BigLep: @SgtPooki : a couple of thoughts:
1. Do you want to paste in the error observed to aid in understanding or debugging?
2. assuming getting this merged is blocking other work you're trying to do for the js-kubo-rpc-client, I think it's safe to merge and then we fast-followup if there's something deeper to address that @achingbrain calls out.

SgtPooki: pushing up a change removing this to see if CI still passes.

PR #777, opened 2022-09-30
feat: convert to typescript

achingbrain:
Converts this module to typescript


src/endpoint/routes.ts
```diff
@@ -77,25 +62,25 @@ export default (server, createFactory) => {
...
         // @ts-expect-error opts is a json object
         nodes[id] = await ipfsd.spawn(opts)
         return {
           id: id,
-          apiAddr: nodes[id].apiAddr ? nodes[id].apiAddr.toString() : '',
-          gatewayAddr: nodes[id].gatewayAddr ? nodes[id].gatewayAddr.toString() : '',
-          grpcAddr: nodes[id].grpcAddr ? nodes[id].grpcAddr.toString() : '',
+          apiAddr: nodes[id].apiAddr?.toString(),
+          gatewayAddr: nodes[id].gatewayAddr?.toString(),
+          grpcAddr: nodes[id].grpcAddr?.toString(),
```
SgtPooki, line 75: we're changing the results from being `''` (empty string) when undefined, to being `undefined`. Is this okay?

src/endpoint/routes.ts
```diff
@@ -137,11 +122,11 @@ export default (server, createFactory) => {
         await nodes[id].start()
 
         return {
-          apiAddr: nodes[id].apiAddr ? nodes[id].apiAddr.toString() : '',
-          gatewayAddr: nodes[id].gatewayAddr ? nodes[id].gatewayAddr.toString() : '',
-          grpcAddr: nodes[id].grpcAddr ? nodes[id].grpcAddr.toString() : ''
+          apiAddr: nodes[id].apiAddr?.toString(),
+          gatewayAddr: nodes[id].gatewayAddr?.toString(),
+          grpcAddr: nodes[id].grpcAddr?.toString()
```
SgtPooki, line 127: we're changing the results from being `''` (empty string) when undefined, to being `undefined`. Is this okay?

src/utils.ts
```diff
@@ -40,95 +32,86 @@ export const defaultRepo = (type) => {
...
+    if (ipfsOptions.pass != null) {
       args.push('--pass', '"' + ipfsOptions.pass + '"')
     }
 
     if (ipfsOptions.preload != null) {
       args.push('--enable-preload', Boolean(typeof ipfsOptions.preload === 'boolean' ? ipfsOptions.preload : ipfsOptions.preload.enabled).toString())
     }
 
-    if (ipfsOptions.EXPERIMENTAL && ipfsOptions.EXPERIMENTAL.sharding) {
+    if ((ipfsOptions.EXPERIMENTAL != null) && ipfsOptions.EXPERIMENTAL.sharding === true) {
```
SgtPooki, line 101: `if (ipfsOptions.EXPERIMENTAL?.sharding === true) {`

src/utils.ts
```diff
@@ -4,33 +4,25 @@ import fs from 'fs'
 import { logger } from '@libp2p/logger'
 import { nanoid } from 'nanoid'
 import tempWrite from 'temp-write'
+import type { ControllerOptions, IPFSOptions, NodeType } from '.'
```
SgtPooki, line 7: `import type { ControllerOptions, IPFSOptions, NodeType } from './types.js'`

SgtPooki (approved):

src/endpoint/routes.ts
achingbrain: Yes, I think it's a bit better like as these fields will be stripped out on JSON encode now as they're undefined instead of being an empty string which is not a valid string representation of a multiaddr.

src/utils.ts
achingbrain: `types.ts` has been removed, all the types are in `index.ts` now.  They were previously re-exported by declaring `@typedef` jsdocs in `index.js` importing from `types.js` but that's not necessary any more.  This could certainly import from `index.js` instead of `.` though it works as is cos tsc can resolve `'.'` then strips it out as it's a type import.

```suggestion
import type { ControllerOptions, IPFSOptions, NodeType } from './index.js'
```

src/utils.ts
achingbrain: Good catch, this style of guarding is done by the `--fix` flag to the lint command.

```suggestion
    if (ipfsOptions.EXPERIMENTAL?.sharding === true) {
```

src/ipfsd-client.ts
```diff
@@ -1,7 +1,8 @@
-import { multiaddr } from '@multiformats/multiaddr'
+import { Multiaddr, multiaddr } from '@multiformats/multiaddr'
 import http from 'ipfs-utils/src/http.js'
 import mergeOptions from 'merge-options'
 import { logger } from '@libp2p/logger'
+import type { Controller, ControllerOptions, InitOptions, IPFSAPI, PeerData, RemoteState } from '.'
```
achingbrain, line 5:
```suggestion
import type { Controller, ControllerOptions, InitOptions, IPFSAPI, PeerData, RemoteState } from './index.js'
```

SgtPooki (approved):

PR #823, opened 2024-04-18
feat!: remove js-ipfs support

achingbrain:
- Removes support for js-ipfs since it is deprecated
- Renames 'go' type argument to 'kubo'
- Still expects to have other implementations added in future
- Updates all deps to latest versions
- Removes deprecated `ipfs-utils` dep

BREAKING CHANGE: only supports spawning `kubo` daemons


.gitignore
```diff
@@ -1,6 +1,9 @@
 node_modules
+build
 dist
 .docs
 .coverage
+node_modules
```
SgtPooki, line 6: Dupe of line 1

SgtPooki (approved):

SgtPooki:
Approved with expectation that ci will be fixed before merging

PR #828, opened 2024-04-19
fix: enable pubsub

achingbrain:
Adds option for pubsub

### DIFF
--- a/src/kubo/utils.ts
+++ b/src/kubo/utils.ts
@@ -66,6 +66,10 @@ export function buildStartArgs (options: KuboStartOptions): string[] {
     args.push('--offline')
   }
 
+  if (options.pubsub === true) {
+    args.push('---enable-pubsub-experiment')
+  }
+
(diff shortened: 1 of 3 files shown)


src/kubo/utils.ts
```diff
@@ -66,6 +66,10 @@ export function buildStartArgs (options: KuboStartOptions): string[] {
     args.push('--offline')
   }
 
+  if (options.pubsub === true) {
+    args.push('---enable-pubsub-experiment')
```
rmlearney-digicatapult, line 70: 3 dashes rather than 2

achingbrain: There was a test for this, but, well:

tests

Fixed in https://github.com/ipfs/js-ipfsd-ctl/commit/bb5f1a1c8f23415a40de34de80224370cf717157 released in https://github.com/ipfs/js-ipfsd-ctl/releases/tag/v14.0.5

PR #850, opened 2024-11-14
fix: return gateway maddr in .info response

### LINKED ISSUE #831: feat: add back `controller.api.gatewayHost` and `controller.api.gatewayPort` or similar
With the latest update from #823 we no longer have the ability to get the listening ipfs gateway host and port from the instantiated kubo node.

this forces consumers to do handle passing around port/host config passed to `createNode` or parsing config options from the created `KuboNode` using https://github.com/multiformats/js-multiaddr-to-uri

We shouldn't need to do this on the consumer side because ipfsd-ctl and kubo-rpc-client should both have the gateway URL readily accessible. 

We should expose this on the `await node.info()` result.

Note: this used to be done by reading stdout: https://github.com/ipfs/js-ipfsd-ctl/blob/6e0d0b5fe49084d38085adfe59a12f19ec8d8c1e/src/ipfsd-daemon.ts#L250

SgtPooki:
In order to not re-add the multiaddr dependency, we're just returning the gateway response as a string. Users will need to parse the string with `@multiformats/multiaddr` or other means.

Fixes #831

### DIFF
--- a/src/kubo/daemon.ts
+++ b/src/kubo/daemon.ts
@@ -27,6 +27,7 @@ export default class KuboDaemon implements KuboNode {
   private readonly disposable: boolean
   private subprocess?: ResultPromise
   private _api?: KuboRPCClient
+  private _gateway?: string
   private readonly repo: string
   private readonly stdout: Logger
   private readonly stderr: Logger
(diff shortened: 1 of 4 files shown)


SgtPooki (commented):
self review

PR #857, opened 2025-04-10
fix: new node should update existing config

SgtPooki:
Currently in draft mode because this PR is just showing a failing test currently.

We should be updating the config with the passed config if the repo already exists.


SgtPooki:
@achingbrain do you have any qualms with the expectation that the config should be updated when a repo already exists?

SgtPooki (commented):
self review

achingbrain:
This seems dangerous if people are using this module to run persistent nodes.

We are/should only be using this module for tests, right? If so why not use a disposable node so the repo is temporary and created each time?

